PR #22, opened 2019-04-05
Fixed admin hooks are not invoked for cron jobs.

fabianmarz:
### Description
- When cron jobs are run, the registered plugin admin hooks are not invoked which prevents some desired behaviour if, for example, posts are updated.
- To prevent this, we call the admin init method manually when doing cron.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -63,6 +63,16 @@ class Plugin {
     // Displays sale price as regular price if custom field is checked.
     add_filter('woocommerce_get_price_html', __NAMESPACE__ . '\WooCommerce::woocommerce_get_price_html', 10, 2);
 
+    // Ensures new product are saved before updating its meta data.
+    add_action('woocommerce_process_product_meta', __NAMESPACE__ . '\WooCommerce::saveNewProductBeforeMetaUpdate', 1);
+    // Updates product delivery time with the lowest delivery time between its own variations.
+    add_action('updated_post_meta', __NAMESPACE__ . '\Admin::updateProductDeliveryTime', 10, 3);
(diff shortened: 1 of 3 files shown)


fabianmarz:
@sun, as this is pending quite a long time, do you think we can merge this even if the desired result is not confirmed? The changes are very straightforward.

src/Plugin.php
```diff
@@ -63,6 +63,16 @@ public static function init() {
...
 
+    // Ensures new product are saved before updating its meta data.
+    add_action('woocommerce_process_product_meta', __NAMESPACE__ . '\WooCommerce::saveNewProductBeforeMetaUpdate', 1);
+    // Updates product delivery time with the lowest delivery time between its own variations.
+    add_action('updated_post_meta', __NAMESPACE__ . '\Admin::updateProductDeliveryTime', 10, 3);
+    // Updates sale percentage when regular price or sale price are updated.
+    add_action('updated_post_meta', __NAMESPACE__ . '\Admin::updateSalePercentage', 10, 4);
+
+    add_action('woocommerce_process_product_meta', __NAMESPACE__ . '\WooCommerce::woocommerce_process_product_meta');
+    add_action('woocommerce_save_product_variation', __NAMESPACE__ . '\WooCommerce::woocommerce_save_product_variation', 10, 2);
```
sun: It's not clear what these two are doing; they were removed from their original context. Seems like they are about product notes and possibly something else?

fabianmarz: Those two hooks are likely to not get triggered at all as they require `$_POST` data, but IMO that's ok. If you want to know what they are doing you can have a look into the method docs.

PR #23, opened 2019-04-10
Fixed shop page is not shown on uppercase containing slugs.

fabianmarz:
### Description
- When using a capital character when trying to access the shop page, the "shop" rewrite rule does not match but the more generic WP one which will display a page rather the shop listing
- This PR is "correcting" this behaviour by checking for capital characters and redirect to the actual shop page if any


src/Plugin.php
```diff
@@ -110,6 +112,28 @@ public static function init() {
...
+   */
+  public static function parse_request($request) {
+    if (!$pagename = $request->query_vars['pagename'] ?? '') {
+      return;
+    };
+    $shop_page = get_post(get_option('woocommerce_shop_page_id'));
+    $shop_page_slug = $shop_page->post_name;
+    // Check if the requested pagename matches the shop page post name
+    // but does not contain case sensitive characters to avoid redirect loops.
+    if (stripos($pagename, $shop_page_slug) !== FALSE && preg_match("@[A-Z]@", $pagename)) {
```
sun:
```suggestion
    if (stripos($pagename, $shop_page_slug) !== FALSE && strpos($pagename, $shop_page_slug) === FALSE) {
```

src/Plugin.php
```diff
@@ -110,6 +112,28 @@ public static function init() {
...
+   * case-insensitive but match a generic page name instead.
+   *
+   * @implements parse_request
+   */
+  public static function parse_request($request) {
+    if (!$pagename = $request->query_vars['pagename'] ?? '') {
+      return;
+    };
+    $shop_page = get_post(get_option('woocommerce_shop_page_id'));
+    $shop_page_slug = $shop_page->post_name;
```
sun:
```suggestion
    $shop_page_slug = get_post_field('post_name', get_option('woocommerce_shop_page_id'), 'raw');
```

sun: Note that this may not be configured yet when the plugin is installed initially / when woocommerce is installed only afterwards.

fabianmarz: Good catch, thank you!

src/Plugin.php
```diff
@@ -110,6 +112,28 @@ public static function init() {
...
+   * This is necessary as the archives rewrite rules do not match
+   * case-insensitive but match a generic page name instead.
+   *
+   * @implements parse_request
+   */
+  public static function parse_request($request) {
+    if (!$pagename = $request->query_vars['pagename'] ?? '') {
+      return;
+    };
+    $shop_page = get_post(get_option('woocommerce_shop_page_id'));
```
sun:
```suggestion
```

src/Plugin.php
```diff
@@ -110,6 +112,28 @@ public static function init() {
     add_action('wp_enqueue_scripts', __CLASS__ . '::wp_enqueue_scripts');
   }
 
+  /**
+   * Redirects uppercase containing shop slugs to the correct shop page.
```
sun:
```suggestion
   * Redirects requests to shop page using different letter-casing to canonical shop page path.
```

src/Plugin.php
fabianmarz: 🙈

PR #24, opened 2019-04-10
Fixed price custom label was displayed for related and cross-sell products.

juanlopez4691:
### Ticket
- [Remove custom price labels from related products](https://app.asana.com/0/383242129844224/1117001017270597)

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -456,12 +456,37 @@ class WooCommerce {
         $price = wc_price(wc_get_price_to_display($product)) . $product->get_price_suffix();
       }
 
-      if (is_product()) {
+      if (is_product() && !static::isSideProduct()) {
         $price_label = get_post_meta($product_id, '_' . Plugin::PREFIX . '_price_label', TRUE) ?: __('(Our price)', Plugin::L10N);
         $price .= ' ' . $price_label;
       }
     }
     return $price;
+  }
+
+  /**
+   * Checks if displayed product is a related, cross-sell or up-sell product.
+   *
+   * @return bool
+   *   TRUE is current displayed product is related, cross-sell or up-sell.
+   */
+  public static function isSideProduct() {
+    global $woocommerce_loop;
+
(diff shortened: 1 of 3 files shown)


src/WooCommerce.php
```diff
@@ -433,6 +433,8 @@ public static function woocommerce_sale_flash($output, $post, $product) {
    * @implements woocommerce_get_price_html
    */
   public static function woocommerce_get_price_html($price, $product) {
+    global $woocommerce_loop;
```
LucaPipolo: This global is not used.

src/WooCommerce.php
```diff
@@ -456,14 +458,34 @@ public static function woocommerce_get_price_html($price, $product) {
         $price = wc_price(wc_get_price_to_display($product)) . $product->get_price_suffix();
       }
 
-      if (is_product()) {
+      if (is_product() && !static::isSideProduct()) {
         $price_label = get_post_meta($product_id, '_' . Plugin::PREFIX . '_price_label', TRUE) ?: __('(Our price)', Plugin::L10N);
-        $price .= ' ' . $price_label;
+        $price .= ' ' . apply_filters(Plugin::PREFIX . '_product_price_label', $price_label);
```
LucaPipolo: Do not like the idea to add a filter if we are not going to use it.

src/WooCommerce.php
```diff
@@ -456,14 +458,34 @@ public static function woocommerce_get_price_html($price, $product) {
...
+
+    if (isset($woocommerce_loop['loop']) && $woocommerce_loop['loop'] === 1) {
+      return FALSE;
+    }
+
+    if (isset($woocommerce_loop['name']) && in_array($woocommerce_loop['name'], ['related', 'cross-sells', 'up-sells'])) {
+      return TRUE;
+    }
+
+    return FALSE;
```
LucaPipolo: Please add some comments to the previous conditions.

PR #25, opened 2019-04-16
Fixed disabled second checkout page allowed repeated orders to be created.

juanlopez4691:
### Ticket
- [Fix order button multiple click prevention](https://app.asana.com/0/383242129844224/1117050697493689)


assets/scripts/main.js
```diff
@@ -5,14 +5,23 @@
...
    * To prevent this we stop form submit event propagation after the first click.
+   *
+   * Covers the case of woocommerce-german-market second checkout page being
+   * disabled. If form fields validation fails, the order submit button is
+   * enabled again to allow the user to resend the form.
    */
-  $('form.checkout.wgm-second-checkout').submit((e) => {
-    if ($('#place_order').hasClass('disabled')) {
+
+  $('#place_order').click(function disableSubmitOrder(e) {
```
sun: It's possible that the unspecified selector might conflict with the multi-step-checkout that was built for STIM.  However, as the website did not go live, we may ignore that possible problem and wait for the case to reappear.

assets/scripts/main.js
```diff
@@ -5,14 +5,23 @@
    * If the page takes time to be loaded, the user could click multiple
    * times on the place order button and this would generate multiple orders.
    * To prevent this we stop form submit event propagation after the first click.
+   *
+   * Covers the case of woocommerce-german-market second checkout page being
+   * disabled. If form fields validation fails, the order submit button is
+   * enabled again to allow the user to resend the form.
```
sun: This addition is more confusing than clarifying now, as both cases are covered with the same code.  Please add any missing info into the first paragraph instead.

LucaPipolo: I agree with Daniel. I do not think we should change the comment at all but, if it needs some improvments, should have been done in the first paragraph.

assets/scripts/main.js
LucaPipolo: @juanlopez4691 — STIM is not live and `shop-standards` is not installed since it was created before STIM. Please, just ignore STIM case.

assets/scripts/main.js
LucaPipolo: I agree with Daniel. I think the comment was pretty claer and do not need modification at all. If you want to add something, I'd do this in the first paragraph.

assets/scripts/main.js
LucaPipolo: @juanlopez4691 — STIM is not live and `shop-standards` plugin is not installed there. Please, proceed ignoring STIM case.

PR #29, opened 2019-05-13
Refactored code.

fabianmarz:
### Ticket
- See base branch

### Description
- Refactored code to not rely on widget class overrides but use DOM traversal to achieve the desired changes instead

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -241,6 +241,48 @@ class Plugin {
   }
 
   /**
+   * Adds the passed argument as query parameter to a set of list elements.
+   *
+   * @param string $content
+   *   The HTML filter list to perform the transformation on.
+   * @param string $arg
+   *   Use delivery time as default argument.
+   *
+   * @return \DOMDocument|null
+   */
+  public static function transformFilterLinks(string $content, $arg = 'delivery_time'): ?\DOMDocument {
+    $args = [];
+    if ($arg = $_GET[$arg] ?? '') {
+      $args = array_filter(array_map('absint', explode(',', wp_unslash($arg))));
+    }
+    // Return early if no deliver time filter is not active.
+    if (!$args) {
+      return NULL;
+    }
+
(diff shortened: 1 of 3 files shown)


src/Plugin.php
```diff
@@ -240,6 +240,48 @@ public static function wpseo_disable_adjacent_rel_links() {
...
+    }
+
+    $doc = new \DOMDocument();
+    $charset = get_option('blog_charset');
+    $options = LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD;
+    $doc->loadHTML(mb_convert_encoding($content, 'HTML-ENTITIES', $charset), $options);
+    $xpath = new \DOMXPath($doc);
+
+    // Select all filter links and add the query parameter to it.
+    $anchors = $xpath->query('//ul/li/a');
```
sun: Not clear why we need to limit to lists?

```suggestion
    $anchors = $xpath->query('//a');
```

src/Plugin.php
```diff
@@ -240,6 +240,48 @@ public static function wpseo_disable_adjacent_rel_links() {
...
+    }
+    // Return early if no deliver time filter is not active.
+    if (!$args) {
+      return NULL;
+    }
+
+    $doc = new \DOMDocument();
+    $charset = get_option('blog_charset');
+    $options = LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD;
+    $doc->loadHTML(mb_convert_encoding($content, 'HTML-ENTITIES', $charset), $options);
```
sun: Why a DOMDocument and not a simple and fast regex?

As the HTML is controlled, a `preg_replace_callback()` would seem to be the most efficient solution, allowing you to control the behavior granularly for each match.

fabianmarz: Pushed a commit which uses regex instead of DOMDocument.

PR #31, opened 2019-05-16
Fix/optimize product titles email 1123159956543922 joan

juanlopez4691:
### Ticket
- [Optimize product titles in order confirmation mail](https://app.asana.com/0/383242129844224/1123159956543922)

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -114,8 +114,7 @@ class Plugin {
       remove_filter('woocommerce_order_item_name', ['WGM_Template', 'add_delivery_time_to_product_title']);
 
       // Removes SKU from order item name, added by woocommerce-german-market.
-      add_action('wgm_email_after_item_name', __NAMESPACE__ . '\WooCommerce::wgm_email_after_item_name', 10, 1);
-      add_action('woocommerce_order_item_meta_start', __NAMESPACE__ . '\WooCommerce::woocommerce_order_item_meta_start', 10, 4);
+      add_filter('woocommerce_email_order_items_args', __NAMESPACE__ . '\WooCommerce::woocommerce_email_order_items_args');
     }
 
     // Enqueues plugin scripts.
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -108,6 +108,16 @@ class Plugin {
     // Adds product attributes to order emails.
     add_filter('woocommerce_display_item_meta', __NAMESPACE__ . '\WooCommerce::woocommerce_display_item_meta', 10, 3);
 
+    if (class_exists('Woocommerce_German_Market')) {
+      // Remove delivery time from product name in order emails, added by
+      // woocommerce-german-market.
+      remove_filter('woocommerce_order_item_name', ['WGM_Template', 'add_delivery_time_to_product_title']);
+
+      // Removes SKU from order item name, added by woocommerce-german-market.
+      add_action('wgm_email_after_item_name', __NAMESPACE__ . '\WooCommerce::wgm_email_after_item_name', 10, 1);
+      add_action('woocommerce_order_item_meta_start', __NAMESPACE__ . '\WooCommerce::woocommerce_order_item_meta_start', 10, 4);
+    }
+
     // Enqueues plugin scripts.
     add_action('wp_enqueue_scripts', __CLASS__ . '::wp_enqueue_scripts');
   }
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -562,6 +562,24 @@ class WooCommerce {
   }
 
   /**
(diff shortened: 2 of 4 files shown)


src/WooCommerce.php
```diff
@@ -561,6 +561,24 @@ public static function woocommerce_display_item_meta($html, $item, $args) {
...
+    ob_start();
+  }
+
+  /**
+   * Removes SKU from order item name, added by woocommerce-german-market.
+   *
+   * @implements woocommerce_order_item_meta_start
+   */
+  public static function woocommerce_order_item_meta_start($item_id, $item, $order, $plain_text) {
+    echo preg_replace('@\s*\(#.+\)@', '', ob_get_clean());
```
sun: If installed on a completely different setup, this could match false positives. Doesn't WGM also add this in a filter/action hook that can be unregistered as well?

juanlopez4691: > If installed on a completely different setup, this could match false positives.
I am checking for WGM to be present before setting up the filter chain. See https://github.com/netzstrategen/wordpress-shop-standards/blob/f2cf3d2d07d9e1786e4e8606eef69b4ae4a516ca/src/Plugin.php#L111. Do you think this might be not safe enough?

> Doesn't WGM also add this in a filter/action hook that can be unregistered as well?
Not that I have seen. Adding SKU to item name seems to be controlled by a WGM backend setting, but:
1. I was not able to find a UI control for that setting in the backend plugin settings
2. The setting seems not to be read before using it in the "New Order" email template.
2. This setting effects are not limited to the case we are trying to fix. Changing it would modify other results. I.e. PDF refund documents.

sun: The templates seem to conditionally output the SKU based on `show_sku`:
https://github.com/netzstrategen/gaco/blob/df10d723c57ed566b4038b528b507b6616ffff6d/wp-content/plugins/woocommerce/templates/emails/email-order-items.php#L53-L55

`show_sku` is passed from:
https://github.com/netzstrategen/gaco/blob/858888251d7f559c19be82b5bbf4a0cdd7548279/wp-content/plugins/woocommerce-german-market/templates/woocommerce-german-market/emails/plain/customer-confirm-order.php#L40
https://github.com/netzstrategen/gaco/blob/858888251d7f559c19be82b5bbf4a0cdd7548279/wp-content/plugins/woocommerce-german-market/templates/woocommerce-german-market/emails/customer-confirm-order.php#L51

This invokes the filter hook `woocommerce_email_order_items_args` in
https://github.com/netzstrategen/gaco/blob/858888251d7f559c19be82b5bbf4a0cdd7548279/wp-content/plugins/woocommerce/includes/wc-template-functions.php#L2993

Isn't it sufficient to set that parameter to false under the desired conditions/scenarios?

juanlopez4691: @sun Thanks a lot, I missed that. I will fix the approach considering your suggestions.

juanlopez4691:
@sun Pushed a fix following your suggestions. Thanks a lot for your detailed review.

src/Plugin.php
```diff
@@ -108,6 +108,15 @@ public static function init() {
     // Adds product attributes to order emails.
     add_filter('woocommerce_display_item_meta', __NAMESPACE__ . '\WooCommerce::woocommerce_display_item_meta', 10, 3);
 
+    if (class_exists('Woocommerce_German_Market')) {
```
sun: 1. The condition seems unnecessary/unwanted, as the variable `show_sku` is also used by WooCommerce itself:
https://github.com/netzstrategen/gaco/blob/858888251d7f559c19be82b5bbf4a0cdd7548279/wp-content/plugins/woocommerce/templates/emails/plain/email-order-details.php#L27

1. I would recommend to move the newly added lines directly below/above the (second to last) filter hook `woocommerce_get_item_data` (line 106) above, so that customizations on the same aspect are grouped together.

juanlopez4691: > The condition seems unnecessary/unwanted, as the variable show_sku is also used by WooCommerce itself...

@sun It seems to me that WooCommerce is using this to add the SKU to the top-over-the-header (before the order table) of notification emails. We only want to remove SKU from item names in the order items table, which is added there by WGM.

> I would recommend to move the newly added lines directly below/above the (second to last) filter hook woocommerce_get_item_data (line 106) above, so that customizations on the same aspect are grouped together.

I will do so.

sun: That's the same spot as WGM; both are targeting the list of order items. WGM is overriding the WooCommerce template. WooCommerce outputs the SKU by default. WGM adds the delivery time to the output.

juanlopez4691: @sun My bad. Closed the PR before seeing your last comment 🙈 You are totally right here. I will add those fixes and create a new PR.

PR #33, opened 2019-05-21
Fixed order of item attributes in confirmation mail.

fabianmarz:
### Ticket
- [Related Ticket](https://app.asana.com/0/30156186242982/1123159956543924/f)

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -534,6 +534,15 @@ class WooCommerce {
 
   /**
    * Adds basic information (e.g. weight, sku, etc.) and product attributes to order emails.
+   * 
+   * @param string $html
+   *  Rendered list of product meta.
+   * 
+   * @param object $item
+   *  The processed item object.
+   *
+   * @param array $args
+   *   The array of formatting arguments.
    *
    * @implements woocommerce_display_item_meta
    */
@@ -541,10 +550,12 @@ class WooCommerce {
     $strings = [];
     $product = $item->get_product();
     $data = static::getProductData($product);
+    // @todo The separator element is stripped from the sent mail. 
     $data[] = [
       'name' => '',
       'value' => '<hr>',
     ];
+    // Add product meta which is contained in $html after product data.
     foreach ($item->get_formatted_meta_data() as $meta_id => $meta) {
       $data[] = [
         'name' => $meta->display_key,
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -541,7 +541,7 @@ class WooCommerce {
     $strings = [];
     $data = [];
     $product = $item->get_product();
-    // Extract key/value pairs from passed html and add it to the product data set.
+    // Extract label/value pairs from passed HTML and add it to the product data set.
     $matched = preg_match_all('@<strong.+[^>]>(.+)</strong>.+<p>(.+)</p>@', $html, $matches, PREG_SET_ORDER);
     if ($matched) {
       foreach ($matches as $match) {
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -534,10 +534,10 @@ class WooCommerce {
 
   /**
    * Adds basic information (e.g. weight, sku, etc.) and product attributes to order emails.
-   * 
+   *
    * @param string $html
    *  Rendered list of product meta.
-   * 
+   *
    * @param object $item
    *  The processed item object.
    *
@@ -550,7 +550,7 @@ class WooCommerce {
     $strings = [];
     $product = $item->get_product();
     $data = static::getProductData($product);
-    // @todo The separator element is stripped from the sent mail. 
+    // @todo The separator element is stripped from the sent mail.
     $data[] = [
       'name' => '',
       'value' => '<hr>',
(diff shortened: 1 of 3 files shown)


src/WooCommerce.php
```diff
@@ -539,8 +539,19 @@ public static function woocommerce_get_item_data($data, $cartItem) {
    */
   public static function woocommerce_display_item_meta($html, $item, $args) {
     $strings = [];
+    $data = [];
     $product = $item->get_product();
-    $product_data_set = array_merge(static::getProductData($product), static::getProductAttributes($product));
+    // Extract key/value pairs from passed html and add it to the product data set.
```
sun:
```suggestion
    // Extract label/value pairs from passed HTML and add it to the product data set.
```

src/WooCommerce.php
```diff
@@ -539,8 +539,19 @@ public static function woocommerce_get_item_data($data, $cartItem) {
...
+    $matched = preg_match_all('@<strong.+[^>]>(.+)</strong>.+<p>(.+)</p>@', $html, $matches, PREG_SET_ORDER);
+    if ($matched) {
+      foreach ($matches as $match) {
+        $data[] = [
+          'name' => rtrim($match[1], ':'),
+          'value' => $match[2],
```
sun: Would it be easier to extract and remove all order item meta data values within the `woocommerce_order_item_meta` hook (or similar name) and temporarily attach them on each `$item`, so that the incoming `$html` is empty and you can output the original/raw meta data here instead of parsing it from the HTML?

fabianmarz: I first thought of this as well, but don't know why I did not do it in the end 🙈. Added a fixup commit, thanks!

fabianmarz:
Note: I also added a separator line, as this is shown in the checkout as well. But currently the <hr> is stripped from the mail template and an empty list element is rendered. Will ask how critical this.

src/WooCommerce.php
```diff
@@ -539,8 +539,19 @@ public static function woocommerce_get_item_data($data, $cartItem) {
    */
   public static function woocommerce_display_item_meta($html, $item, $args) {
     $strings = [];
+    $data = [];
```
sun: The separator could be added first to reduce the amount things being juggled here (and juggle them in their actual order).

```suggestion
    $data = [];
    // Add a separator before product attributes.
    $data[] = [
      'name' => '',
      'value' => '<hr>',
    ];
```

src/WooCommerce.php
```diff
@@ -552,11 +563,20 @@ public static function woocommerce_display_item_meta($html, $item, $args) {
...
+        $string .= $productData['value'];
+      }
+      if (!empty($string)) {
+        $strings[] = $string;
+      }
     }
 
     if ($strings) {
-      $html = $args['before'] . implode($args['separator'], $strings) . $args['after'] . $html;
+      $html = $args['before'] . implode($args['separator'], $strings) . $args['after'];
```
sun: What's contained in the original `$html`?  Wouldn't it contain output from other plugins that gets lost by removing it?

src/WooCommerce.php
```diff
@@ -552,11 +563,20 @@ public static function woocommerce_display_item_meta($html, $item, $args) {
     }
 
     foreach ($product_data_set as $productData) {
-      $strings[] = '<strong class="wc-item-meta-label">' . $productData['name'] . ':</strong> ' . $productData['value'];
+      $string = '';
+      if (!empty($productData['name'])) {
```
sun: Does it make sense to output a heading if there is no value?

Shouldn't it rather be negated?  If there is no heading, skip to the next item?

src/WooCommerce.php
```diff
@@ -552,11 +563,20 @@ public static function woocommerce_display_item_meta($html, $item, $args) {
...
+        $string .= '<strong class="wc-item-meta-label">' . $productData['name'] . ':</strong> ';
+      }
+      if (!empty($productData['value'])) {
```
sun: What if the meta data value is zero? (`"0"`)

```suggestion
      if (isset($productData['value']) && strlen(trim($productData['value']))) {
```

src/WooCommerce.php
sun: It also seems like you could easily get rid of the multidimensional array_merge:

```suggestion
    $data = static::getProductData($product);
    // Add a separator before product data and attributes.
    $data[] = [
      'name' => '',
      'value' => '<hr>',
    ];
```

The same probably applies to the delivery time splicing in between, although I still did not really understand yet into which exact position it is sliced into.

src/WooCommerce.php
fabianmarz: The `$html` is coming from the [`wc_display_item_meta()`](https://github.com/netzstrategen/gaco/blob/master/wp-content/plugins/woocommerce/includes/wc-template-functions.php#L3018-L3047) function and basically calls the same code I do here (`get_formatted_meta_data()`) to create the html string. So this should not contain any other information.

src/WooCommerce.php
fabianmarz: I introduced this for the separator, as I want to display a <hr> without any preceding value. But after checking, the hr is stripped and an empty <li> is rendered. But as we spent quite some time already, I decided to leave it and address this at a later point.

src/WooCommerce.php
fabianmarz: 👍

PR #28, opened 2019-05-10
Added delivery time filter to product listing.

juanlopez4691:
### Ticket
- [WGM - filter by "delivery time"](https://app.asana.com/0/383242129844224/1119011613980983)

### Details
Finally solved this adding a custom URL query var (`delivery_time`), modifying the WP query (`pre_get_post`), creating a new widget (`WidgetFilterDeliveryTime`) and overriding two widgets from woocommerce (`WC_Widget_Layered_Nav` and `WidgetLayeredNavFilters`).

PR #35, opened 2019-05-24
Fixed taxonomy query var might be NULL.

fabianmarz:
### Description
- Regression fix for #28

### DIFF
--- a/src/ProductFilters/WidgetFilterDeliveryTime.php
+++ b/src/ProductFilters/WidgetFilterDeliveryTime.php
@@ -154,7 +154,7 @@ class WidgetFilterDeliveryTime extends \WC_Widget {
    *
    * @implements get_terms_orderby
    */
-  public static function get_terms_orderby(string $orderby, array $query_vars, ?array $taxonomy): string {
+  public static function get_terms_orderby(string $orderby, array $query_vars, array $taxonomy = []): string {
     if ($taxonomy && in_array('product_delivery_times', $taxonomy, TRUE) && $query_vars['orderby'] === 'slug') {
       $orderby = $orderby . '+0';
     }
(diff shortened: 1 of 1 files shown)


src/ProductFilters/WidgetFilterDeliveryTime.php
```diff
@@ -154,8 +154,8 @@ public static function getProductsDeliveryTimes(array $args = []) {
    *
    * @implements get_terms_orderby
    */
-  public static function get_terms_orderby(string $orderby, array $query_vars, array $taxonomy): string {
-    if (in_array('product_delivery_times', $taxonomy, TRUE) && $query_vars['orderby'] === 'slug') {
+  public static function get_terms_orderby(string $orderby, array $query_vars, ?array $taxonomy): string {
```
sun: Wasn't aware that optional function parameters were introduced already — and actually can't find it in the PHP manual.  Do you have a link?

No mention of this no any of these:
https://www.php.net/manual/en/migration70.new-features.php
https://www.php.net/manual/en/migration71.new-features.php
https://www.php.net/manual/en/migration72.new-features.php
https://www.php.net/manual/en/migration73.new-features.php

sun:
```suggestion
  public static function get_terms_orderby(string $orderby, array $query_vars, array $taxonomy = []): string {
```

juanlopez4691: @sun It's the first "new feature" for PHP 7.1 https://www.php.net/manual/en/migration71.new-features.php

fabianmarz: As stated by @juanlopez4691, to cite the docs:
> Type declarations for parameters and return values can now be marked as nullable by prefixing the type name with a question mark.

PR #36, opened 2019-05-27
Fixed taxonomy might be type NULL.

fabianmarz:
### Description
- 🙈 sorry… committed untested code. This PR fixes the issue.

### DIFF
--- a/src/ProductFilters/WidgetFilterDeliveryTime.php
+++ b/src/ProductFilters/WidgetFilterDeliveryTime.php
@@ -154,7 +154,7 @@ class WidgetFilterDeliveryTime extends \WC_Widget {
    *
    * @implements get_terms_orderby
    */
-  public static function get_terms_orderby(string $orderby, array $query_vars, array $taxonomy = []): string {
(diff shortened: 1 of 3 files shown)


src/ProductFilters/WidgetFilterDeliveryTime.php
```diff
@@ -154,7 +154,7 @@ public static function getProductsDeliveryTimes(array $args = []) {
    *
    * @implements get_terms_orderby
    */
-  public static function get_terms_orderby(string $orderby, array $query_vars, array $taxonomy = []): string {
+  public static function get_terms_orderby(string $orderby, array $query_vars, ?array $taxonomy): string {
```
sun: From https://www.php.net/manual/en/migration71.new-features.php:

The nullable type declaration only works if the calling code explicitly passes `NULL`, otherwise it throws:
> Uncaught Error: Too few arguments to function test(), 2 passed in...

Is this your intention?

```suggestion
  public static function get_terms_orderby(string $orderby, array $query_vars, array $taxonomy = []): string {
```

PR #37, opened 2019-05-29
Added automatic selection of product variation attributes when only one option available.

juanlopez4691:
### Ticket
- [Benutzerfreundliche Variantenwahl / user-friendly selection of variations](https://app.asana.com/0/383242129844224/1120121728917861)


juanlopez4691:
> `else` should be on a new line.

Linter was complaining. I will put it in a new line, then.

PR #40, opened 2019-08-19
Fixed sale price as regular price removes energy label.

fabianmarz:
### Ticket
- [Related Ticket](https://app.asana.com/0/784106262441860/1136103500287444/f) (internal only)

### Description
- When checking "Display sale price as normal price" it removes other components which were added via the `woocommerce_get_price_html` filter as the price is getting overridden

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -463,12 +463,12 @@ class WooCommerce {
         }
         $price = wc_price(wc_get_price_to_display($product)) . $product->get_price_suffix();
       }
+      // Invoke regular price_html filters (excluding this one).
       $current_filter = current_filter();
-      $priority = has_filter($current_filter, __CLASS__ . '::woocommerce_get_price_html');
-      // Remove and re-add filter to avoid callback loop.
-      remove_filter($current_filter, __CLASS__ . '::woocommerce_get_price_html', $priority);
+      $priority = has_filter($current_filter, __METHOD__);
+      remove_filter($current_filter, __METHOD__);
       $price = apply_filters($current_filter, $price, $product);
-      add_filter($current_filter, __CLASS__ . '::woocommerce_get_price_html', $priority, 2);
+      add_filter($current_filter, __METHOD__, $priority, 2);
 
       if (is_product() && !static::isSideProduct()) {
         $price_label = get_post_meta($product_id, '_' . Plugin::PREFIX . '_price_label', TRUE) ?: __('(Our price)', Plugin::L10N);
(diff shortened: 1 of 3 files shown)


src/WooCommerce.php
```diff
@@ -456,12 +456,20 @@ public static function woocommerce_get_price_html($price, $product) {
...
       }
       else {
         if (!$product->get_sale_price()) {
           return $price;
         }
         $price = wc_price(wc_get_price_to_display($product)) . $product->get_price_suffix();
+        // Remove and re-add filter to avoid callback loop.
+        remove_filter('woocommerce_get_price_html', __CLASS__ . '::woocommerce_get_price_html');
+        $price = apply_filters('woocommerce_get_price_html', $price, $product);
+        add_filter('woocommerce_get_price_html', __CLASS__ . '::woocommerce_get_price_html', 10, 2);
```
sun: 1. To prevent further incompatibilities in the future, we should retrieve the actual priority of the already registered filter and re-register it with the same priority.

1. You can DRY by moving the lines below the current condition, so it is executed in both cases.

fabianmarz: 2. makes total sense. Sorry for missing this 🙈😅

src/WooCommerce.php
```diff
@@ -463,6 +463,12 @@ public static function woocommerce_get_price_html($price, $product) {
         }
         $price = wc_price(wc_get_price_to_display($product)) . $product->get_price_suffix();
       }
+      $current_filter = current_filter();
+      $priority = has_filter($current_filter, __CLASS__ . '::woocommerce_get_price_html');
+      // Remove and re-add filter to avoid callback loop.
+      remove_filter($current_filter, __CLASS__ . '::woocommerce_get_price_html');
```
sun: https://developer.wordpress.org/reference/functions/remove_filter/ states
> To remove a hook, the $function_to_remove and $priority arguments must match when the hook was added. This goes for both filters and actions. No warning will be given on removal failure.

The default value is 10, but this will fail if the priority was changed.
```suggestion
      remove_filter($current_filter, __CLASS__ . '::woocommerce_get_price_html', $priority);
```

I also wonder whether the callback function name isn't the same as `__FUNCTION__` in this context, but that's a very minor detail.

Similarly minor, the inline comment should be above the block of lines, not in between, and should explain what is being done; e.g.:
```php
// Invoke regular price_html filters (excluding this one).
```

fabianmarz: Thanks for catching! Overseen this as well.

fabianmarz: Added a fixup using `__METHOD__` as this returns the full path while `__FUNCTION__` returns the name only.

PR #41, opened 2019-08-20
Added salutation field to billing and shipping address.

fabianmarz:
### Ticket
- [Related Ticket](https://app.asana.com/0/30156186242982/1122978077654777/f) (internal only)

### Description
- This PR adds an optional salutation field to the shipping and billing checkout fields
- The field can be enabled via the backend
- I also moved the already existing settings to a dedicated tab "Shop Standards" Tab page to have them grouped together


src/WooCommerce.php
```diff
@@ -721,4 +721,26 @@ public static function getProductAttributes(\WC_Product $product) {
     return $data;
   }
 
+  /**
+   * Add salutation field for billing and shipping.
+   *
+   * @implements woocommerce_checkout_fields
+   */
+  public static function woocommerce_checkout_fields(?array $fields): array {
```
sun: I can't remember all details, but I'm fairly sure that this should use the hook `woocommerce_default_address_fields` as in https://github.com/netzstrategen/wordpress-ssofact/blob/master/src/WooCommerce.php#L302-L314

src/WooCommerce.php
```diff
@@ -721,4 +721,26 @@ public static function getProductAttributes(\WC_Product $product) {
...
+    $fields['shipping']['shipping_salutation'] = [
+      'label' => __('Salutation', Plugin::L10N),
+      'class' => ['form-row-wide'],
+      'type' => 'select',
+      'options' => [
+        __('Mr', Plugin::L10N) => __('Mr', Plugin::L10N),
+        __('Ms', Plugin::L10N) => __('Ms', Plugin::L10N),
+        __('Company', Plugin::L10N) => __('Company', Plugin::L10N),
```
sun: If we are implementing this as a standard in 2019, then I think there should be at least an option "Diverse" or alternatively "Prefer not to say" here.

src/Plugin.php
```diff
@@ -143,6 +143,11 @@ public static function init() {
       add_filter('woocommerce_email_styles', __NAMESPACE__ . '\WooCommerce::woocommerce_email_styles');
     }
 
+    if (get_option('_add_salutation_field') === 'yes') {
```
sun: Missing plugin namespace prefix.

src/WooCommerce.php
fabianmarz: This code is copied over from https://github.com/netzstrategen/woocommerce-moeve/blob/master/src/Plugin.php#L61 and works fine there.

src/WooCommerce.php
fabianmarz: We discussed this and decided against for now.

src/Plugin.php
fabianmarz: While I agree, I oriented on the already existing option names which don't use a prefix as well. If we add a prefix, we should add it to the other options as well.

fabianmarz:
Creating a dedicated plugin for it sounds bit of overkill as it's about adding a single field only. In addition, the plugin holds lots of other features which should probably be extracted into dedicated classes as well, no? IMO we can start extracting the salutation field first.

src/WooCommerce.php
fabianmarz: Forgot to mention, this is an optional field, which brings me to the point of adding a null value.

fabianmarz:
@sun, I came up with a new approach in https://github.com/netzstrategen/wordpress-shop-standards/pull/41/commits/181d1748befb68b6e269a6891e44a1aa19ecb66b where I started to move code into related classes. Please have a look again 🙂

src/Checkout.php
```diff
@@ -0,0 +1,78 @@
...
+ * @file
+ * Contains \Netzstrategen\ShopStandards\Checkout.
+ */
+
+namespace Netzstrategen\ShopStandards;
+
+/**
+ * Checkout related settings and actions.
+ */
+class Checkout {
```
sun: The proposed structure looks good, but I'd strongly recommend to scope the classes into features and thus have a class for Salutation only.

If anything generic remains, we can keep a generic Checkout class for that.

src/Checkout.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  /**
+   * Adds checkout specific settings.
+   *
+   * @implements woocommerce_get_settings_shop_standards
+   */
+  public static function woocommerce_get_settings_shop_standards(array $settings): array {
+    $settings = array_merge($settings, [
```
sun: It would be simpler to use regular array additions instead of a merge here:
```php
$settings[] = [
  'type' => 'title',
  'name' => ...
];
$settings[] = [
  'type' => 'checkbox',
  ...
];
```
I'd also recommend to always state the type first. Makes it easier to understand what each code block represents.

src/Plugin.php
```diff
@@ -88,6 +88,9 @@ public static function init() {
     add_action('woocommerce_process_product_meta', __NAMESPACE__ . '\WooCommerce::woocommerce_process_product_meta');
     add_action('woocommerce_save_product_variation', __NAMESPACE__ . '\WooCommerce::woocommerce_save_product_variation', 10, 2);
 
+    Seo::init();
```
fabianmarz: I did not rename this as it's not tightly coupled to WooCommerce.

src/WooCommerce.php
sun: If I remember correctly, it will not appear in the administrative user profile or somewhere else this way. But okay, I guess this can still be refactored later, if necessary.

sun:
Please squash-merge using the corrected PR title.

marcbinz:
@fabianmarz Is it correct, that the salutation mapping makes use of the translated value both for key and for value?

```
'options' => [
    __('Mr', Plugin::L10N) => __('Mr', Plugin::L10N),
    __('Ms', Plugin::L10N) => __('Ms', Plugin::L10N),
    __('Company', Plugin::L10N) => __('Company', Plugin::L10N),
],
```

AFAICT, the normal (untranslated) value should be stored in the database and only be translated for display purposes. If this is already the case and I am just not seeing it, I am sorry 😬 
When passing the values to Moeve, I need to map from the String value to the Integer one (as this is what Moeve expects), so I would appreciate the untranslated version so I can use a constant map for that 😉

PR #46, opened 2019-09-23
Added formal translations.

fabianmarz:
### Description
- This PR adds a missing formal translations file which is basically a copy of the regular file.


sun:
@fabianmarz What do you mean with "basically a copy"? Is there any change?

If there isn't, then we don't need these copies. Formal German always needs to be used with the plugin https://wordpress.org/plugins/language-fallback/

fabianmarz:
Thanks for pointing out. Sorry I forgot about this 🙈 PR can be closed as there are no content related changes in the translation.

PR #49, opened 2019-10-10
Fixed scripts synchronous loading was decreasing site pages performance.

juanlopez4691:
### Ticket
- [Defer or dequeue WooCommerce assets on non-WooCommerce pages](https://app.asana.com/0/383242129844224/1141315041731476)

### Description
-

### DIFF
--- a/src/Performance.php
+++ b/src/Performance.php
@@ -10,9 +10,9 @@ class Performance {
   /**
    * WooCommerce scripts to be defer or async loaded.
    *
-   * @var string
+   * @var array
    */
-  const SCRIPTS_LOAD = [
+  const SCRIPTS_ASYNC_LOAD = [
     'jquery-blockui' => 'defer',
     'wc-add-to-cart' => 'defer',
     'js-cookie' => 'defer',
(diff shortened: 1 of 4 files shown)


src/Plugin.php
```diff
@@ -26,6 +26,29 @@ class Plugin {
    */
   private static $baseUrl;
 
+  /**
+   * Scripts to be defer loaded.
```
sun: Can we move this new code into new class named "Performance", please?

src/Plugin.php
```diff
@@ -220,6 +246,27 @@ public static function register_acf() {
...
+   */
+  public static function script_loader_tag($tag) {
+    $loading_types = [];
+    $loading_types['defer'] = apply_filters(Plugin::L10N . '/scripts_load_defer', static::SCRIPTS_LOAD_DEFER);
+    $loading_types['async'] = apply_filters(Plugin::L10N . '/scripts_load_async', static::SCRIPTS_LOAD_ASYNC);
+
+    foreach ($loading_types as $attr => $scripts) {
+      foreach ($scripts as $script) {
+        if (strpos($tag, $script) !== FALSE) {
+          $tag = str_replace(' src=', sprintf(' %s="%s" src=', $attr, $attr), $tag);
```
sun: HTML5, please.

```suggestion
          $tag = str_replace(' src=', sprintf(' %s src=', $attr), $tag);
```

src/Plugin.php
```diff
@@ -220,6 +246,27 @@ public static function register_acf() {
...
 
+  /**
+   * Loads scripts as deferred or async.
+   *
+   * @implements script_loader_tag
+   */
+  public static function script_loader_tag($tag) {
+    $loading_types = [];
+    $loading_types['defer'] = apply_filters(Plugin::L10N . '/scripts_load_defer', static::SCRIPTS_LOAD_DEFER);
+    $loading_types['async'] = apply_filters(Plugin::L10N . '/scripts_load_async', static::SCRIPTS_LOAD_ASYNC);
```
sun: This architecture is quite confusing. I first thought that these would be two different processing staging, but it turns out that they are just two separate stacks per attribute.  I would suggest the following instead:

1. Use the script handles instead of the filesystem paths, because the latter can change more often. The handle is passed to the filter; see https://developer.wordpress.org/reference/hooks/script_loader_tag/

1. Use only one constant to specify both async and defer. Use the array key for the handle, and explicit value items to adjust the attributes.

```php
  const SCRIPTS = [
    'wc-add-to-cart' => ['async' => FALSE, 'defer' => TRUE],
    'wc-add-to-cart-variation' => ['async' => FALSE, 'defer' => TRUE],
  ];
```

juanlopez4691:
@sun Pushed a fix.

juanlopez4691:
@sun Fixed conflicts. Can you please review last fixes?

src/Performance.php
```diff
@@ -0,0 +1,214 @@
...
+ * Performance improvements.
+ */
+class Performance {
+
+  /**
+   * WooCommerce scripts to be defer or async loaded.
+   *
+   * @var array
+   */
+  const SCRIPTS_ASYNC_LOAD = [
```
fabianmarz: Please sort A-Z. 

Why did you decide against the specification of both `async` and `defer` as suggest by @sun in https://github.com/netzstrategen/wordpress-shop-standards/pull/49#discussion_r333558972?

Additionally a good short read: https://flaviocopes.com/javascript-async-defer/#async-and-defer

src/Performance.php
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  /**
+   * Gets URLs of resources to prefetch.
+   *
+   * @return array
+   *   URLs of resources to prefetch.
+   */
+  public static function getResourcesUrlPrefetch() {
+    return apply_filters(Plugin::L10N . '/esources_url_prefetch', static::RESOURCES_URL_PREFETCH);
```
fabianmarz:
```suggestion
    return apply_filters(Plugin::L10N . '/resources_url_prefetch', static::RESOURCES_URL_PREFETCH);
```

src/Performance.php
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  /**
+   * Gets styles handles to dequeue.
+   *
+   * @return array
+   *   Handles of styles to be dequeued.
+   */
+  public static function getDequeueStyles(): array {
+    return apply_filters(Plugin::L10N . '/styles_dequeue', static::STYLES_DEQUEUE);
```
fabianmarz: The filter should match the method name (and elsewhere). IMO it's more clear if you put the verb before the noun (`dequeue_styles`, `async_load_styles`, `async_load_scripts`). Another option would be to group scripts and styles together 

```
styles_async
styles_dequeue
```

```suggestion
    return apply_filters(Plugin::L10N . '/dequeue_styles', static::DEQUEUE_STYLES);
```

src/Plugin.php
```diff
@@ -227,7 +247,6 @@ public static function register_acf() {
       'active' => 1,
     ]);
   }
-
```
fabianmarz: Empty line should not be removed.

src/Performance.php
juanlopez4691: Ok, I assumed `defer` and `async` modes where mutually excluding.

LucaPipolo:
All changes requested by @sun and @fabianmarz seems to be correctly applied.
I did not detected other necessary fixes.

@juanlopez4691, please solve the version conflict and feel free to proceed.

PR #52, opened 2019-10-24
Fixed backend setting to control display of sale label was missing.

juanlopez4691:
### Ticket
- []()

### Description
- Backend setting to define a minimum discount percentage to display the sale bubble was missing.


src/WooCommerce.php
```diff
@@ -26,6 +33,20 @@ public static function woocommerce_get_settings_shop_standards(array $settings):
...
     ];
+    $settings[] = [
+      'name' => __('Products settings', Plugin::L10N),
+      'type' => 'title',
+    ];
+    $settings[] = [
+      'id' => '_minimum_sale_percentage_to_display_label',
+      'type' => 'text',
+      'name' => __('Minimum discount percentage to display product sale label', Plugin::L10N),
+      'default' => 10,
```
fabianmarz: Should be `static::SALE_BUBBLE_MIN_AMOUNT` as well, or?

src/WooCommerce.php
```diff
@@ -448,7 +469,7 @@ public static function woocommerce_sale_flash($output, $post, $product) {
     else {
       $sale_percentage = get_post_meta($product->get_id(), '_sale_percentage', TRUE);
     }
-    $minimum_sale_percentage = get_option('_minimum_sale_percentage_to_display_label', 10);
+    $minimum_sale_percentage = get_option('_minimum_sale_percentage_to_display_label', SALE_BUBBLE_MIN_AMOUNT);
```
fabianmarz: As this is a class constant it should be `static::SALE_BUBBLE_MIN_AMOUNT`
```suggestion
    $minimum_sale_percentage = get_option('_minimum_sale_percentage_to_display_label', static::SALE_BUBBLE_MIN_AMOUNT);
```

fabianmarz:
Code looks good to me. Did you deploy it somewhere so it can be tested?

src/WooCommerce.php
juanlopez4691: Ah! Anyway I will fix that.

juanlopez4691:
@fabianmarz This is now deployed on http://feco.stage.netzstrategen.com

assets/scripts/main.js
```diff
@@ -46,4 +48,57 @@
...
+      $singleProductSaleLabel.show();
+    } else {
+      $singleProductSaleLabel.hide();
+    }
+  }
+
+  $('.single_variation_wrap')
+    .on('show_variation', (event, variation) => {
+      if (variation.display_price < variation.display_regular_price) {
+        // eslint-disable-next-line max-len
```
colourgarden: Can we move this onto the line itself and use `// eslint-disable-line max-len`? Would be a bit neater and wouldn't interrupt the reading flow.

assets/scripts/main.js
```diff
@@ -46,4 +48,57 @@
...
     }
   });
+
+  // Single product sale label DOM element.
+  const $singleProductSaleLabel = $('.single-product-summary .onsale');
+
+  /**
+   * Hides sale label if percentage value is lower than 10%.
+   */
+  $('.onsale').each(function hideSaleLabel() {
```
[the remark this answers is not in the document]
LucaPipolo: AFAIK, anonymous function are not allowed by our ESLint standards (inerithed from Google).

PR #53, opened 2019-10-24
Added validation to ensure that GTIN is unique.

moritzbappert:
### Ticket
- [Prevent GTIN duplicates](https://app.asana.com/0/587433704548192/1112402955414747)

### Description
- This is not quite done yet but I would like to get feedback on whether you guys think it's a good way I chose to tackle this.
- GTIN is checked for uniqueness via AJAX on change of the field
- an error message is shown below the field if duplicate GTINs are found
- saving the product is not prevented, post meta won't be updated if GTIN isn't unique though

Any ideas on better styling? What is missing to provide a solid enough solution?


moritzbappert:
Note to myself: I still need to check whether variable products are already covered.

moritzbappert:
Hey @sun, thanks for your feedback! 

> Instant feedback in the form is nice, but you still need to prevent a product from being saved with a duplicate GTIN. The instant feedback only improves UX to prevent other values from not being saved due to a late validation error. A programmatically saved product does not execute JS. You can test the programmatic saving with WP-CLI `wp post save ...`.

I will figure out a way to prevent saving the post via PHP and to issue an error message (which is more effort than one might expect), the downside is nothing gets saved then. 

I have forgotten to include some code in my commits, that at least doesn't update post meta to a duplicate GTIN - so this is a partial solution, but error messages are still missing there anyway.
Regardless of this, is *saving the post via PHP and issuing an error message* your preferred implementation?

src/Plugin.php
```diff
@@ -88,9 +88,15 @@ public static function init() {
...
     add_filter('woocommerce_get_settings_shop_standards', __NAMESPACE__ . '\WooCommerce::woocommerce_get_settings_shop_standards');
 
+    // Allow ajax requests to specified functions.
+    add_action('wp_ajax_is_existing_gtin', __NAMESPACE__ . '\WooCommerce::ajax_is_existing_gtin');
+
+    // Enqueues admin plugin scripts.
+    add_action('admin_enqueue_scripts', __CLASS__ . '::admin_enqueue_scripts');
+
     Seo::init();
     WooCommerceSalutation::init();
```
moritzbappert: Just to understand: what is this for (in comparison to the distinction between `Plugin.php` and   `Admin.php`)?

sun:
> Regardless of this, is **saving the post via PHP and issuing an error message** your preferred implementation?

I would keep the instant validation via JS to increase the chance of valid data before saving. The post/product should be saved normally though, not by your code. I would expect a pre-save validation that cancels the save process in case the data is not valid.

src/Plugin.php
sun: `Plugin::init()` is invoked at all times. Everything below the `return is_admin()` in there is only executed in the front-end, not in the administration interface.

`Admin::init()` is only invoked in the administration interface.

moritzbappert:
> I would expect a pre-save validation that cancels the save process in case the data is not valid.

I'll try that. The only thing I've seen before was a reload of the post edit screen upon clicking the save button with an added error message (probably there's no other way without JS).

src/Plugin.php
moritzbappert: https://github.com/netzstrategen/wordpress-shop-standards/blob/299f984cc92c6920fc41e2724356ee27ef41642a/src/Plugin.php#L81-L89	

I don't want to open up a huge discussion here, but wouldn't lines like these need to be placed in `Admin.php` as well? They are never needed in the front-end imho.

moritzbappert:
Hey @sun,
I have now been able to easily add error messages through `\WC_Admin_Meta_Boxes::add_error()`. 
When a post is saved through WP-CLI though, no error gets shown:
update-product-with-existing-gtin
The already existing GTIN isn't saved, but it seems like everything went fine.
For improving this, one would probably need to hook to `save_post`. I don't think it's efficient having me trying to solve this last bit though.

moritzbappert:
@sun anything you want to say about this?

assets/scripts/admin.js
```diff
@@ -0,0 +1,31 @@
+(function pageLoad($) {
+  /**
+   * Checks GTIN for uniqueness and issues an error message if necessary.
+   */
+  $(document).on('change', "input[name*='_shop-standards_gtin']", (event) => {
+    const $gtinInput = $(event.currentTarget);
+    const $gtinField = $gtinInput.parents('.form-field');
+    $gtinField.find('.error').remove();
+    // TODO: Show loading spinner.
```
LucaPipolo: We should add this, shouldn't we?

assets/scripts/admin.js
```diff
@@ -0,0 +1,31 @@
...
+      function (response) {
+        if (!response.is_unique) {
+          const message = shop_standards_admin.gtin_error_message.replace('{{url}}', response.duplicate_edit_link);
+          $gtinField.append(`
+            <div class="error error-message">
+              ${message}
+            </div>
+          `);
+        }
+        // TODO: Success message for unique GTIN.
```
LucaPipolo: We should add this, shouldn't we?

languages/shop-standards-de_DE.po
```diff
@@ -248,6 +276,13 @@ msgstr "Herr"
 msgid "Company"
 msgstr "Firma"
 
+#~ msgid ""
```
LucaPipolo: Please, remove all these commented lines.

src/WooCommerce.php
```diff
@@ -223,7 +223,14 @@ public static function woocommerce_process_product_meta($post_id) {
     foreach ($custom_fields as $field) {
       if (isset($_POST[$field])) {
         if (!is_array($_POST[$field]) && $_POST[$field]) {
-          update_post_meta($post_id, $field, $_POST[$field]);
+          if ($field !== '_' . Plugin::PREFIX . '_gtin') {
+            update_post_meta($post_id, $field, $_POST[$field]);
+          } else if (!self::is_existing_gtin($post_id, $_POST[$field])) {
```
LucaPipolo: We are following [Drupal Code Standards for control structures](https://www.drupal.org/docs/develop/standards/coding-standards#controlstruct). Therefore, the `else if` should be on a new line.

src/WooCommerce.php
```diff
@@ -223,7 +223,14 @@ public static function woocommerce_process_product_meta($post_id) {
...
+            // GTIN is only saved if unique.
+            update_post_meta($post_id, $field, $_POST[$field]);
+          } else {
```
LucaPipolo: Same here, please move the else on a new line.

src/WooCommerce.php
```diff
@@ -380,6 +387,7 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
       return;
     }
     $variation_id = $_POST['variable_post_id'][$loop];
+    $gtin_error = false;
```
LucaPipolo: We are following [Drupal Code Standards for constant naming](https://www.drupal.org/docs/develop/standards/coding-standards#s-constants). So `true` should be uppercase.

src/WooCommerce.php
```diff
@@ -389,14 +397,25 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
     foreach ($custom_fields as $field) {
       if (isset($_POST[$field]) && isset($_POST[$field][$loop])) {
         if ($_POST[$field][$loop]) {
-          update_post_meta($variation_id, $field, $_POST[$field][$loop]);
+          if ($field !== '_' . Plugin::PREFIX . '_gtin') {
+            update_post_meta($post_id, $field, $_POST[$field]);
+          } else if (!self::is_existing_gtin($variation_id, $_POST[$field][$loop])) {
```
LucaPipolo: Same here, please move the `else if` on a new line.

src/WooCommerce.php
```diff
@@ -389,14 +397,25 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
...
+            // GTIN is only saved if unique.
+            update_post_meta($variation_id, $field, $_POST[$field][$loop]);
+          } else {
```
LucaPipolo: Same here, please move the `else` on a new line.

src/WooCommerce.php
```diff
@@ -223,7 +223,14 @@ public static function woocommerce_process_product_meta($post_id) {
...
         if (!is_array($_POST[$field]) && $_POST[$field]) {
-          update_post_meta($post_id, $field, $_POST[$field]);
+          if ($field !== '_' . Plugin::PREFIX . '_gtin') {
+            update_post_meta($post_id, $field, $_POST[$field]);
+          } else if (!self::is_existing_gtin($post_id, $_POST[$field])) {
+            // GTIN is only saved if unique.
+            update_post_meta($post_id, $field, $_POST[$field]);
+          } else {
+            \WC_Admin_Meta_Boxes::add_error(__('The entered GTIN isn\'t unique and couldn\'t be saved. Please enter a unique GTIN and save again. All other changes have been saved.', Plugin::L10N));
+          }
```
LucaPipolo: I'd propose to refactoring the conditional logic to ensure the error message is displayed only in case the field is GTIN and the GTIN is existing.

Something like:
```
if ($field !== '_' . Plugin::PREFIX . '_gtin') {
  update_post_meta($post_id, $field, $_POST[$field]);
}
else {
  if (!self::is_existing_gtin($post_id, $_POST[$field])) {
    // update
  }
  else {
    // error
  }
}
```

src/WooCommerce.php
```diff
@@ -389,14 +397,25 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
...
         if ($_POST[$field][$loop]) {
-          update_post_meta($variation_id, $field, $_POST[$field][$loop]);
+          if ($field !== '_' . Plugin::PREFIX . '_gtin') {
+            update_post_meta($post_id, $field, $_POST[$field]);
+          } else if (!self::is_existing_gtin($variation_id, $_POST[$field][$loop])) {
+            // GTIN is only saved if unique.
+            update_post_meta($variation_id, $field, $_POST[$field][$loop]);
+          } else {
+            $gtin_error = true;
+          }
```
LucaPipolo: I'd propose to refactoring the conditional logic to ensure the GTIN error message is set only in case the field is GTIN and the GTIN is existing.

Something like what I proposed above.

src/WooCommerce.php
```diff
@@ -389,14 +397,25 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
...
       if (isset($_POST[$field]) && isset($_POST[$field][$loop])) {
         if ($_POST[$field][$loop]) {
-          update_post_meta($variation_id, $field, $_POST[$field][$loop]);
+          if ($field !== '_' . Plugin::PREFIX . '_gtin') {
+            update_post_meta($post_id, $field, $_POST[$field]);
+          } else if (!self::is_existing_gtin($variation_id, $_POST[$field][$loop])) {
+            // GTIN is only saved if unique.
+            update_post_meta($variation_id, $field, $_POST[$field][$loop]);
+          } else {
+            $gtin_error = true;
```
LucaPipolo: Also here `true` should be uppercase.

src/WooCommerce.php
```diff
@@ -389,14 +397,25 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
...
         }
         else {
           delete_post_meta($variation_id, $field);
         }
       }
     }
 
+    if ($gtin_error) {
+      \WC_Admin_Meta_Boxes::add_error(__('At least one entered GTIN in product variations isn\'t unique and couldn\'t be saved. Please enter unique GTINs and save again. All other changes have been saved.', Plugin::L10N));
+    }
```
LucaPipolo: Can not understand why we are displaying the error twice with a different message. 

I mean, I think we could extract the error message into a function and call it from `woocommerce_process_product_meta ` and `woocommerce_save_product_variation ` but using always the same message. 

If do so, please ensure to clean as well the related translation.

src/WooCommerce.php
```diff
@@ -750,4 +769,58 @@ public static function getProductAttributes(\WC_Product $product) {
     return $data;
   }
 
+  /**
+   * Check if product gtin is found for any other product IDs.
```
LucaPipolo:
```suggestion
   * Checks if product GTIN is found for any other product IDs.
```

src/WooCommerce.php
```diff
@@ -750,4 +769,58 @@ public static function getProductAttributes(\WC_Product $product) {
...
+   *
+   * @param int $product_id Product ID.
+   * @param int $gtin GTIN.
+   * @return bool
+   */
```
LucaPipolo: The PHP DocBlock should be formulated like this:

```
/**
 * Description of the method.
 *
 * @param type name
 *   Description of the parameter.
 *
 * @return type
 *   Description of the return.
 */

Notice that the description of parameters and return are on the next line and indented with 3 spaces.
As commented above, the method description should use the third person.

File: src/WooCommerce.php [hunk 16]
```diff
@@ -750,4 +769,58 @@ public static function getProductAttributes(\WC_Product $product) {
...
+    );
+  }
+
+  /**
+   * Wrap is_existing_gtin for response to ajax requests.
+   *
+   * @param int $product_id Product ID.
+   * @param int $gtin GTIN.
+   * @return bool
+   */
```
LucaPipolo: See my previous comment about PHP DocBlock.

src/WooCommerce.php
```diff
@@ -750,4 +769,58 @@ public static function getProductAttributes(\WC_Product $product) {
...
+  public static function ajax_is_existing_gtin() {
+    $product_id = $_POST['product_id'];
+    $gtin = $_POST['gtin'];
+    $response = array(
+      'is_unique' => true,
+    );
```
LucaPipolo: We are using [Drupal Code Standards to write arrays](https://www.drupal.org/docs/develop/standards/coding-standards#array).

So it should be:

```
$response = [
  'is_unique' => TRUE,
];
```

src/WooCommerce.php
```diff
@@ -750,4 +769,58 @@ public static function getProductAttributes(\WC_Product $product) {
...
+    if ($found_duplicate = self::is_existing_gtin($product_id, $gtin)) {
+      // For product variations, we want to link to the edit page of the parent product.
+      $found_duplicate = get_post_type($found_duplicate) === 'product_variation' ? wp_get_post_parent_id($found_duplicate) : $found_duplicate;
+      $duplicate_edit_link = get_edit_post_link($found_duplicate);
+
+      $response = array(
+        'is_unique' => false,
+        'found_duplicate' => $found_duplicate,
+        'duplicate_edit_link' => $duplicate_edit_link,
+      );
```
LucaPipolo: Same thing commented above. `[]` instead of `array()`.

moritzbappert:
@LucaPipolo, thanks for all the helpful comments! I'll improve the code and will update you here. :)

moritzbappert:
Hey @LucaPipolo,
I have addressed all your comments and have pushed changes as two fixup commits.
Afterwards I did `git rebase origin/master`, and then force-pushed the changes, as I thought this was the right thing to solve the conflicts. It doesn't seem so though. Can you please advice me how to proceed?

assets/scripts/admin.js
```diff
@@ -0,0 +1,44 @@
...
+          $gtinField.append(`
+            <div class="notice notice-success" style="margin: 15px 0 5px">
+              ${message}
+            </div>
+          `);
+        }
+        else {
+          const message = shop_standards_admin.gtin_error_message.replace('{{url}}', response.duplicate_edit_link);
+          $gtinField.append(`
+            <div class="notice notice-error" style="margin: 15px 0 5px">
```
LucaPipolo: Could we extract the CSS into a style file instead of mixing it in the JS?

languages/shop-standards-de_DE.po
```diff
@@ -148,31 +161,40 @@ msgstr ""
...
-#: src/WooCommerce.php:341
+#: src/WooCommerce.php:352
 msgid "ERP/Inventory ID:"
 msgstr "ERP/Inventar ID:"
 
-#: src/WooCommerce.php:357
+#: src/WooCommerce.php:368
 msgid "Variation has insufficient images"
-msgstr "Variation hat ein unzureichendes Bilder"
+msgstr "Variation hat unzureichende Bilder"
```
LucaPipolo: Is this changes intended? To me, seems not related with the current feature. 🤔

languages/shop-standards-de_DE.po
```diff
@@ -239,17 +261,9 @@ msgid "Mr"
 msgstr "Herr"
 
 #: src/WooCommerceSalutation.php:65
-#, fuzzy
-#| msgid "Mr"
 msgid "Mrs"
-msgstr "Herr"
+msgstr "Frau"
```
LucaPipolo: ditto

src/Admin.php
```diff
@@ -28,9 +28,38 @@ public static function init() {
...
+  public static function admin_enqueue_scripts($hook) {
+    $git_version = Plugin::getGitVersion();
+    global $post;
+    if ($hook == 'post-new.php' || $hook == 'post.php') {
+      if ($post->post_type === 'product') {
+        wp_enqueue_script(Plugin::PREFIX . '_admin', Plugin::getBaseUrl() . '/dist/scripts/admin.min.js', ['jquery'], $git_version, TRUE);
+        wp_localize_script(Plugin::PREFIX . '_admin', 'shop_standards_admin', array(
+          'product_id' => $post->ID,
+          'gtin_error_message' => sprintf(
+            __('The entered GTIN <a href="%s">exists already</a>. It must be changed in order to save the product.', Plugin::L10N),
```
LucaPipolo: I do not understand why the localized error message is different from the message shown on `WooCommerce.php:235` and `WooCommerce.php:413`. Can we not extract all the message errors into a single method and apply the same translation to all of them? 🤔

src/WooCommerce.php
```diff
@@ -386,10 +397,22 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
...
+          if ($field !== '_' . Plugin::PREFIX . '_gtin') {
+            update_post_meta($variation_id, $field, $_POST[$field][$loop]);
+          }
+          else {
+            if (!self::is_existing_gtin($variation_id, $_POST[$field][$loop])) {
+              // GTIN is only saved if unique.
+              update_post_meta($variation_id, $field, $_POST[$field][$loop]);
+            }
+            else {
+              \WC_Admin_Meta_Boxes::add_error(__('The entered GTIN isn\'t unique and couldn\'t be saved. Please enter a unique GTIN and save again. All other changes have been saved.', Plugin::L10N));
```
LucaPipolo: This code looks duplicated to me. Line 235 of this file is doing the same. Can we extract this into a function? Would be easier to manage/change the translation in that way.

src/WooCommerce.php
```diff
@@ -750,4 +773,58 @@ public static function getProductAttributes(\WC_Product $product) {
     return $data;
   }
 
+  /**
+   * Checks if product GTIN is found for any other product IDs.
+   *
+   * @param int $product_id
+   *   Product ID.
```
LucaPipolo:
```suggestion
   *   The product ID.
```

src/WooCommerce.php
```diff
@@ -750,4 +773,58 @@ public static function getProductAttributes(\WC_Product $product) {
...
+   *
+   * @param int $gtin
+   *   GTIN.
```
LucaPipolo:
```suggestion
   *   The GTIN.
```

src/WooCommerce.php
```diff
@@ -750,4 +773,58 @@ public static function getProductAttributes(\WC_Product $product) {
...
+   *
+   * @return bool
```
LucaPipolo: Still missing the return description on the next line.
Also [`get_var` method](https://developer.wordpress.org/reference/classes/wpdb/get_var/#return) seems to return a string or `NULL`, not `bool`.

src/Admin.php
```diff
@@ -28,9 +28,38 @@ public static function init() {
     // Adds products variations custom fields.
     add_action('woocommerce_product_after_variable_attributes', __NAMESPACE__ . '\WooCommerce::woocommerce_product_after_variable_attributes', 10, 3);
 
+    // Allow ajax requests to specified functions.
+    add_action('wp_ajax_is_existing_gtin', __NAMESPACE__ . '\WooCommerce::ajax_is_existing_gtin');
```
LucaPipolo:
```suggestion
    add_action('wp_ajax_is_existing_gtin', __NAMESPACE__ . '\WooCommerce::wp_ajax_is_existing_gtin');
```

src/WooCommerce.php
```diff
@@ -750,4 +773,58 @@ public static function getProductAttributes(\WC_Product $product) {
...
+        $gtin,
+        $product_id
+      )
+    );
+  }
+
+  /**
+   * Wraps is_existing_gtin for responding to ajax requests.
+   */
+  public static function ajax_is_existing_gtin() {
```
LucaPipolo:
```suggestion
  public static function wp_ajax_is_existing_gtin() {
```

src/WooCommerce.php
```diff
@@ -750,4 +773,58 @@ public static function getProductAttributes(\WC_Product $product) {
...
+        '_' . Plugin::PREFIX . '_gtin',
+        $gtin,
+        $product_id
+      )
+    );
+  }
+
+  /**
+   * Wraps is_existing_gtin for responding to ajax requests.
+   */
```
LucaPipolo: Please add an extra comment line with `@implements wp_ajax_is_existing_gtin`. Leave a blank line before it. :)

assets/scripts/admin.js
moritzbappert: I thought it would be a lot of overhead for just one small style improvement – but okay, I'll do it.

languages/shop-standards-de_DE.po
moritzbappert: Not related, but wrong anyway.

src/WooCommerce.php
moritzbappert: Does this suffice?
```   
   * @return string|null
   *   Database query result (post_id of first found duplicate GTIN as string), or null on failure.
```

src/Admin.php
moritzbappert: The context is a little different: this message is used for AJAX validation (before anything is saved), the other one is shown after all other data has been saved – it's only a slight difference, but I still think it makes sense to differentiate. 
I might add the edit link of the product where the duplicate has been found to the notice after server side validation as well though, to add some value. What do you think?

moritzbappert:
Hey @LucaPipolo,
I've solved all the conflicts and have addressed all comments but the one regarding different error messages - I'm waiting for your reply on this.

src/Admin.php
LucaPipolo: @moritzbappert — still does not agree about the difference between AJAX and non-AJAX messages. But I will leave this choice up to you.

Anyway, yes: agree about the edit link of the product, if possible.

Also, please adjust the message:

```diff
- __('The entered GTIN <a href="%s">exists already</a>. It must be changed in order to save the product.', Plugin::L10N),
+ __('The entered GTIN <a href="%s">already exists</a>. It must be changed in order to save the product.', Plugin::L10N),
```

moritzbappert:
Thanks, @LucaPipolo! 
I've adjusted the error message, added the edit link and have decided to unify both error messages into one. ;) 

Now only a review by @sun is outstanding to make this mergeable.

LucaPipolo:
@moritzbappert — PR approved but please, ⚠️ bump the version on `plugin.php` and `package.json` files before merge!

moritzbappert:
@LucaPipolo I have bumped the plugin version to 1.31.0.

src/Admin.php
```diff
@@ -30,6 +30,38 @@ public static function init() {
...
+
+  /**
+   * Enqueues admin plugin scripts.
+   *
+   * @implements admin_enqueue_scripts
+   */
+  public static function admin_enqueue_scripts($hook) {
+    $git_version = Plugin::getGitVersion();
+    global $post;
+    if ($hook == 'post-new.php' || $hook == 'post.php') {
```
LucaPipolo:
```suggestion
    if ($hook === 'post-new.php' || $hook === 'post.php') {
```

sun: @moritzbappert Can you apply this change and/or grant us write access to your fork, please?

moritzbappert: @sun @LucaPipolo I have just applied the suggested changes!

moritzbappert:
I just recognized that there was an error in `Admin.php` with a line of `Settings::init();` which resulted in a fatal error, as `init()` doesn't exist in the Settings class – no idea how it got in there. I've fixed this and force pushed the amended commit.

moritzbappert:
@colourgarden just recognized that I left conflicts in some files. 🙈

PR #55, opened 2019-11-11
Added product brand to schema.org structured data.

LucaPipolo:
### Ticket
- [Missing brand in schema.org on single product pages](https://app.asana.com/0/106767930274882/1141746824096235)

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -150,9 +150,9 @@ class Plugin {
     add_action('wp_enqueue_scripts', __CLASS__ . '::wp_enqueue_scripts');
 
     // Adds GTIN product number in schema.org.
-    add_filter('woocommerce_structured_data_product', __NAMESPACE__ . '\Seo::get_product_gtin');
+    add_filter('woocommerce_structured_data_product', __NAMESPACE__ . '\Seo::getProductGrin');
     // Adds Brand name to schema.org.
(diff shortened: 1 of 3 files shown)


src/Plugin.php
```diff
@@ -150,7 +150,9 @@ public static function init() {
     add_action('wp_enqueue_scripts', __CLASS__ . '::wp_enqueue_scripts');
 
     // Adds GTIN product number in schema.org.
-    add_filter('woocommerce_structured_data_product', __NAMESPACE__ . '\Schema::get_product_gtin');
+    add_filter('woocommerce_structured_data_product', __NAMESPACE__ . '\Seo::get_product_gtin');
+    // Adds Brand name to schema.org.
+    add_filter('woocommerce_structured_data_product', __NAMESPACE__ . '\Seo::get_product_brand');
```
juanlopez4691: Shouldn't those methods names be came case?

src/Seo.php
```diff
@@ -76,4 +76,53 @@ public static function wpseo_disable_adjacent_rel_links(): bool {
...
+        break;
+
+      case 14:
+        $gtin_format_type = 'gtin14';
+        break;
+
+      default:
+    }
+
+    $gtin_format_type = 'gtin12';
```
juanlopez4691: I think this line should meant to be the `default` case of the preceding ` switch-case`block, right?

LucaPipolo: Sorry, it was unintentionally moved during the copy and paste. Good catch.

PR #68, opened 2019-11-28
Added low stock option to product availability text.

colourgarden:
### Ticket
- [Show low stock levels in frontend](https://app.asana.com/0/30156186242982/1133642108266488/f)

### Description
- Comparing `low_stock_amount` (default WooCommerce field) with the current stock level to determine if a product is "low stock"


src/WooCommerce.php
```diff
@@ -119,19 +119,33 @@ public static function woocommerce_ajax_variation_threshold($qty, $product) {
...
    * whether stock managing is enabled or it's available.
    *
    * @implements woocommerce_get_availability
    */
   public static function woocommerce_get_availability($stock, $product) {
     $product->set_manage_stock('yes');
     $product->set_backorders('yes');
+
+    // Low stock level is only set on parent product.
+    // If product variation then manually get the value based on parent ID.
```
juanlopez4691:
```suggestion
    // If product is a variation then manually get the value based on parent ID.
```

PR #69, opened 2019-12-10
Added new checkbox in order to disable energy label displaying per default.

LucaPipolo:
### Ticket
- [Add new checkbox in order to disable energy label displaying per default](https://app.asana.com/0/106767930274882/1151258293585977)


LucaPipolo:
As discussed on the related task, we decided to create our own custom plugin to solve that issue.

The commit in that PR is solving an issue with backend settings not get saved that was already solved by Daniel on https://github.com/netzstrategen/wordpress-shop-standards/pull/73.

Therefore, closing that PR draft.

PR #70, opened 2019-12-11
Fixed incorrect notice when zero stock but backorder possible.

colourgarden:
### Ticket
- [Show low stock levels in frontend](https://app.asana.com/0/30156186242982/1133642108266488/f)

### Description
- Products can be ordered even if stock level is zero.
- Previously, "only 0 in stock" was shown in this scenario.

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -134,10 +134,10 @@ class WooCommerce {
     if ($product->is_type('variation')) {
       $low_stock_amount = get_post_meta($product->get_parent_id(), '_low_stock_amount', TRUE);
     }
-    elseif ($product->get_low_stock_amount()) {
+    else {
       $low_stock_amount = $product->get_low_stock_amount();
     }
-    else {
+    if (!$low_stock_amount) {
       $low_stock_amount = get_option('woocommerce_notify_low_stock_amount');
     }
 
@@ -158,6 +158,7 @@ class WooCommerce {
         $stock['class'] = 'in-stock';
       }
     }
+
     return $stock;
   }
 
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -136,7 +136,7 @@ class WooCommerce {
     else {
       $low_stock_amount = $product->get_low_stock_amount();
     }
-    if ($low_stock_amount === '') {
+    if (empty($low_stock_amount)) {
       $low_stock_amount = get_option('woocommerce_notify_low_stock_amount');
     }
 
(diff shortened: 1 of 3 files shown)


src/WooCommerce.php
```diff
@@ -128,29 +128,35 @@ public static function woocommerce_get_availability($stock, $product) {
...
+    // If low stock threshold is not set at product level, get global option.
     if ($product->is_type('variation')) {
-      $low_stock_amount = get_post_meta($product->get_parent_id(), '_low_stock_amount', TRUE);
+      $parent_product = wc_get_product($product->get_parent_id());
+      $low_stock_amount = $parent_product->get_low_stock_amount();
     }
     else {
       $low_stock_amount = $product->get_low_stock_amount();
     }
+    if ($low_stock_amount === '') {
```
LucaPipolo: Can `$low_stock_amount` be undefined at this point? If yes, we should check it with `isset`.

Also, why not using `empty` PHP method? AFAIK, the only difference is that `empty` will consider a string containing `0` as `TRUE`. Your solution is instead checking only an exactly empty string.

If we should consider `0` as a falsy value, I think we should make it more explicit in the code (by casting) or adding a comment.

src/WooCommerce.php
```diff
@@ -128,29 +128,35 @@ public static function woocommerce_get_availability($stock, $product) {
...
-      elseif ($product->get_stock_quantity() <= $low_stock_amount) {
-        $stock['availability'] = sprintf(__('Only %s in stock', Plugin::L10N), $low_stock_amount);
+    // Check global "Stock display format" option is set to show low stock notices.
+    $show_low_stock_amount = get_option('woocommerce_stock_format') === 'low_amount';
+
+    // is_in_stock() returns true if stock level is zero but backorders allowed.
+    // If stock level below threshold (but above zero), show "Only x in stock".
+    // Otherwise, show "In stock" (not shown by WooCommerce by default).
+    if ($product->is_in_stock()) {
+      if ($show_low_stock_amount && $product->get_stock_quantity() > 0 && $product->get_stock_quantity() <= $low_stock_amount) {
```
LucaPipolo: We are calling three times the `get_stock_quantity` method. To improve performances, we should run it once and save it's result into a variable.

src/WooCommerce.php
colourgarden: I actually copied these lines from WooCommerce Core - https://github.com/woocommerce/woocommerce/blob/master/includes/wc-stock-functions.php#L340-L343

I thought it was better to stick with the Core method but I guess that `empty()` would be better. I'll update 👍

colourgarden: Fixed.

src/WooCommerce.php
colourgarden: Fixed.

src/WooCommerce.php
LucaPipolo: As we discussed in person, we need to not consider `0` value for that conditional.
Therefore, @colourgarden is going to revert the fix and explain in a comment why we are only checking for empty strings.

colourgarden: After discussing offline further, we decided to revert to the WooCommerce empty string check as zero is actually a valid value for the WooCommerce field.

PR #72, opened 2020-01-09
Fixed sale category does not get assigned on product update.

LucaPipolo:
### Ticket
- [Automate sale label (category) assignment](https://app.asana.com/0/inbox/106767930205241/1138362657798199/1156166804867037)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.30.4",
+  "version": "1.30.5",
   "title": "shop-standards",
   "description": "Standard refinements for e-commerce websites.",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.30.3",
+  "version": "1.30.4",
```
sun: The patch version needs to be bumped to .5 now.

src/Plugin.php
```diff
@@ -99,9 +99,14 @@ public static function init() {
     add_action('woocommerce_process_product_meta', __NAMESPACE__ . '\WooCommerce::woocommerce_process_product_meta');
     add_action('woocommerce_save_product_variation', __NAMESPACE__ . '\WooCommerce::woocommerce_save_product_variation', 10, 2);
 
-// Adds woocommerce specific settings.
+    // Adds woocommerce specific settings.
     add_filter('woocommerce_get_settings_shop_standards', __NAMESPACE__ . '\WooCommerce::woocommerce_get_settings_shop_standards');
 
+    // Assigns sale category conditionally on product update.
```
sun: The lines are no longer removed from Admin since the rebase:
https://github.com/netzstrategen/wordpress-shop-standards/blob/a8c648b4c65c44f754dedd527950ef6a95b8bf6a/src/Admin.php#L29-L32

PR #75, opened 2020-01-13
Fixed PHP error thrown in backend when product delivery time is not set.

juanlopez4691:
### Description
- Error reported on Slack by @stephansperling . https://netzstrategen.slack.com/archives/C8RP4J0MR/p1578917736000500

### DIFF
--- a/src/Admin.php
+++ b/src/Admin.php
@@ -78,7 +78,8 @@ class Admin {
       }
     }
     asort($variation_deliveries_ranges);
-    return array_keys($variation_deliveries_ranges)[0];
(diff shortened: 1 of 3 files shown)


src/Admin.php
```diff
@@ -78,7 +78,8 @@ public static function getProductShortestDeliveryTime(\WC_Product $product): str
       }
     }
     asort($variation_deliveries_ranges);
-    return array_keys($variation_deliveries_ranges)[0];
+    $shortest_delivery_time = array_keys($variation_deliveries_ranges)[0] ?: '';
```
sun:
```suggestion
    $shortest_delivery_time = key($variation_deliveries_ranges) ?: '';
```

PR #77, opened 2020-01-24
Fixed sale percentage was not updated for bundle products.

juanlopez4691:
### Ticket
- [Sale Bubble shows wrong value "-0 %" / NUTR & LOFI](1154739415381321)

### Description
-

### DIFF
--- a/src/Admin.php
+++ b/src/Admin.php
@@ -98,7 +98,7 @@ class Admin {
         $parent_id = $product->get_parent_id();
         static::saveSalePercentage($parent_id, get_post($parent_id));
       }
-      else {
+      elseif ($product->get_type() === 'simple' || $product->get_type() === 'bundle') {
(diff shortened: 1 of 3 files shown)


src/Admin.php
```diff
@@ -98,7 +98,7 @@ public static function updateSalePercentage($check, $object_id, $meta_key, $meta
         $parent_id = $product->get_parent_id();
         static::saveSalePercentage($parent_id, get_post($parent_id));
       }
-      elseif ($product->get_type() === 'simple') {
+      else {
```
LucaPipolo: As we discussed in person, I'm a bit scared of removing the condition. There might be more edge cases. 

What if a third party plugin is adding another product type like bundle did? Would be the following statement correct in that case? 

Would not be safer to add an logical or on the condition to include bundles?
Like `elseif ($product->get_type() === 'simple' || $product->get_type() === 'bundle')`.

PR #84, opened 2020-03-13
Fixed variation image update on automated variation selection.

juanlopez4691:
### Ticket
- [Product image does not adjust to selected attribute](https://app.asana.com/0/587433704548192/1164605910938414/f)

### Description
- See https://github.com/netzstrategen/wopa/pull/471


LucaPipolo:
@juanlopez4691 — please, the PR title! 🙏 🙏

assets/scripts/main.js
```diff
@@ -44,9 +43,14 @@
...
         }
-        $variationsForm.trigger('check_variations');
       });
     }
+    // Ensure the rigth product image is displayed.
+    // Some delay seems to be needed to refresh the product image.
+    // We couldn't find a proper event to hook on, so we used a timeout.
+    setTimeout(() => {
+      $variationsForm.trigger('check_variations');
+    }, 100);
```
LucaPipolo: As we discussed during a Slack call, I really do not like this `setTimeout` approach.
To me it looks weird and wrong to rely on it. As you probably know, `setTimeout` is just a call to the browser API and therefore is not called in a synchronous way. Pretty hard to ensure this get always executed at the right moment.

bogdanarizancu: @LucaPipolo, I totally agree, however I still cannot find a better way to do this, since the native woocommerce js code does the same (https://github.com/woocommerce/woocommerce/blob/master/assets/js/frontend/add-to-cart-variation.js#L43-L48). Please let me know if you can come up with something else, thank you.

LucaPipolo: As we discussed on last Slack call, I still think this is a really poor approach on the technical side but it's also true that, as @bogdanarizancu said, WooCommerce is using the same approach. 🙈 

Therefore, I think we can do nothing different from what @bogdanarizancu and  @juanlopez4691 proposed.

PR #89, opened 2020-03-30
Fixed sale percentage was not recalculated after product price update.

juanlopez4691:
### Ticket
- [Sale Bubble not updated when_sale_price_dates_from is set](https://app.asana.com/0/809933051638353/1166600580702113/f)

### Description
-

### DIFF
--- a/src/Admin.php
+++ b/src/Admin.php
@@ -129,7 +129,7 @@ class Admin {
    * @implements updated_post_meta
    */
   public static function updateSalePercentage($check, $object_id, $meta_key, $meta_value) {
-    if ($meta_key === '_regular_price' || $meta_key === '_sale_price') {
+    if (in_array($meta_key, ['price', '_regular_price', '_sale_price'])) {
       $product = wc_get_product($object_id);
       if ($product->get_type() === 'variation') {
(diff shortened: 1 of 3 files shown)


src/Admin.php
```diff
@@ -129,7 +129,7 @@ public static function getProductShortestDeliveryTime(\WC_Product $product): str
    * @implements updated_post_meta
    */
   public static function updateSalePercentage($check, $object_id, $meta_key, $meta_value) {
-    if ($meta_key === '_regular_price' || $meta_key === '_sale_price') {
+    if (in_array($meta_key, ['price', '_regular_price', '_sale_price'])) {
```
LucaPipolo, line 132: As we [discussed on Slack](https://netzstrategen.slack.com/archives/CK598LKEV/p1585905605009500), I'm a bit afraid of triggering the ´updateSalePercentage´ method every time a price changes, even if the product is (or was not) in sale.

But, as I understand from @juanlopez4691 clarification, we can do anything different since when a product sale period ends, WooCommerce is changing the `price` meta key, and not the `_regular_price` or `_sale_price` one.

PR #90, opened 2020-04-03
Fixed sale percentage label was hidden on product page with if no variation is selected.

juanlopez4691:
### Ticket
- [Display sales bubbles on product pages with variants immediately upon page view.]()

### Description
-


plugin.php
```diff
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Shop Standards
-  Version: 1.30.18
+  Version: 1.30.19
```
LucaPipolo, line 5: I'm not sure this is a fix. Sounds more a new minor feature to me.

juanlopez4691: Meh, not sure about that but to me it feels more like a fix to the already existing functionality.

PR #94, opened 2020-04-17
Fixed variable products delivery time was not displayed after woocommerce-german-market update.

juanlopez4691:
### Ticket
- [Fix after update issue: Missing output of the delivery time value for variable products](https://app.asana.com/0/1156709897787098/1171572465854081)

### Description
-


fabianmarz:
@juanlopez4691, as there is a single commit only I don't see which changes were made to make it work. Could you please add a dedicated commit for the template override?

juanlopez4691:
> @juanlopez4691, as there is a single commit only I don't see which changes were made to make it work. Could you please add a dedicated commit for the template override?
@fabianmarz This is not a template override, as @LucaPipolo says, but a class method override. The original method/class lines which I am removing are referenced in the class block comment.

https://github.com/netzstrategen/wopa/blob/4dab5a48216d0fa3e42f85d1107e018fc89a3196/local-plugins/woocommerce-german-market/inc/WGM_Template.php#L2519-L2523

fabianmarz:
In any case it would be easier to make a dedicated commit to the original copied method. And make the changes afterwards. But I'll have a look. Thanks.

src/WGMTemplate.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    // If product is out of stock, don't show delivery time.
+    if (!self::show_delivery_time_if_product_is_not_in_stock($product)) {
+      $lieferzeit_output = apply_filters('gm_delivery_time_message_if_out_of_stock', '', $product);
+    }
+
+    // Output delivery time.
+    if (!empty($lieferzeit_output)) {
+      $lieferzeit_and_markup = '<div class="wgm-info shipping_de shipping_de_string ' . 'delivery-time-' . sanitize_title($label_string) . '">
```
fabianmarz, line 68: I just trying to understand what's going on here. But if we don't need to add the `'delivery-time-' . sanitize_title($label_string) . ` class couldn't we do something like the following:

```php
add_filter('wgm_deliverytime_loop', function ($output, $label) {
  global $product;
  if (!$label) {
    $output .= \WGM_Template::get_deliverytime_string($product);
  }
  return $output;
}, 10, 2);
```

juanlopez4691:
@fabianmarz Checked your suggested fix and works good. I dropped my monstrosity in favor of your clever solution.

PR #96, opened 2020-05-05
Fixed low return products could be purchased.

LucaPipolo:
### Ticket
- ["Plus Products" Prevent WooCommerce Checkout for Add-on Products](https://app.asana.com/0/inbox/106767930205241/1173393553044195/1174246637908210)


src/PlusProducts.php
```diff
@@ -0,0 +1,82 @@
...
+    ];
+    $settings[] = [
+      'type' => 'sectionend',
+      'id' => Plugin::L10N,
+    ];
+    return $settings;
+  }
+
+  /**
+   * Shows an error notice if customer try to buy only Plus products.
```
juanlopez4691, line 47:
```suggestion
   * Shows an error notice if customer tries to buy only Plus products.
```

src/PlusProducts.php
```diff
@@ -0,0 +1,82 @@
...
+   */
+  public static function displayPlusProductsNotice() {
+    $plusCategory = get_term_by('id', get_option('_' . Plugin::L10N . '_plus_products_category'), 'product_cat');
+
+    if (is_wp_error($plusCategory) || empty($plusCategory)) {
+      return;
+    }
+
+    if (static::checkProductsCategoryInCart($plusCategory->slug)) {
+      wc_add_notice(__('Actually, there are only Plus products in your shopping cart. Please, add at least one product from a different category to complete your order.', Plugin::L10N), 'error');
```
juanlopez4691, line 59:
```suggestion
      wc_add_notice(__('Currently there are only Plus products in your shopping cart. Please, add at least one product from a different category to complete your order.', Plugin::L10N), 'error');
```

src/PlusProducts.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  /**
+   * Checks if a cart contains products only from a given category.
```
juanlopez4691, line 64:
```suggestion
   * Checks if cart contains products only from a given category.
```

src/PlusProducts.php
```diff
@@ -0,0 +1,82 @@
...
+   *
+   * @param string $categorySlug
+   *   The slug of the plus products category.
```
juanlopez4691, line 67:
```suggestion
   *   The slug of the products category.
```

src/PlusProducts.php
```diff
@@ -0,0 +1,82 @@
...
+   *
+   * @return bool
+   *   True if the cart only contains the plus products category.
```
juanlopez4691, line 70:
```suggestion
   *   True if the cart contains only products from the given category.
```

src/PlusProducts.php
```diff
@@ -0,0 +1,82 @@
...
+   */
+  public static function checkProductsCategoryInCart(string $categorySlug): bool {
+    foreach (WC()->cart->get_cart() as $cartItem) {
+      if (!has_term($categorySlug, 'product_cat', $cartItem['product_id'])) {
+        return FALSE;
+      }
```
juanlopez4691, line 76: I am not sure about this logic. Products can have more than one category, right? What if the product belongs to the given category (Plus?) but also to some other? Is this situation possible? Makes sense?

LucaPipolo: Good question, @juanlopez4691 . 

As I just clarified with @stephansperling , the correct behaviour would be to stop the cart as soon the only product contained in it belogs **also** to the Plus category. 

So I think the current logic is correct. ✅

PR #97, opened 2020-05-11
Fixed Plus Products error message was shown if the cart was empty.

LucaPipolo:
### Ticket
- [Fix Plus Product warning message is displayed on empty carts](https://app.asana.com/0/106767930274882/1175093376311869)

### DIFF
--- a/src/PlusProducts.php
+++ b/src/PlusProducts.php
@@ -49,8 +49,10 @@ class PlusProducts {
    * @implements woocommerce_check_cart_items
    */
   public static function displayPlusProductsNotice() {
+    if (WC()->cart->is_empty()) {
+      return;
+    }
     $plusCategory = get_term_by('id', get_option('_' . Plugin::L10N . '_plus_products_category'), 'product_cat');
-
     if (is_wp_error($plusCategory) || empty($plusCategory)) {
       return;
     }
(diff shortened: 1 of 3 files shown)


src/PlusProducts.php
```diff
@@ -51,7 +51,7 @@ public static function woocommercePlusProductsSettings(array $settings): array {
   public static function displayPlusProductsNotice() {
     $plusCategory = get_term_by('id', get_option('_' . Plugin::L10N . '_plus_products_category'), 'product_cat');
 
-    if (is_wp_error($plusCategory) || empty($plusCategory)) {
+    if (is_wp_error($plusCategory) || empty($plusCategory) || WC()->cart->is_empty()) {
       return;
     }
```
juanlopez4691, line 56: Why not check first if cart empty and do an early return? Chances are that the category term query is not needed in some cases.
```
if (WC()->cart->is_empty()) {
  return;
}
$plusCategory = get_term_by('id', get_option('_' . Plugin::L10N . '_plus_products_category'), 'product_cat');
if (empty($plusCategory) || is_wp_error($plusCategory)) {
  return;
}

LucaPipolo: Yeah, sure. Good catch, thanks. 👍

PR #99, opened 2020-05-13
Removed sale percentage feature.

LucaPipolo:
### Ticket
- [Fix wrong sale bubble percentage](https://app.asana.com/0/inbox/106767930205241/1175325398085777/1175549924708738)


LucaPipolo:
@fabianmarz — sorry, forgot to describe this properly.

It's true, we kept the [hide sale percentage feature](https://github.com/netzstrategen/wordpress-shop-standards/blob/fix/sale-percentage-refactor-luca/src/WooCommerce.php#L201-L206) in the `shop-standard` plugin, but this was intentional.

The original goal of this feature, was to hide/show the sale label. Does not matter if this is a percentage label or a defaul fixed price label from WooCommerce.

The feature and the related field were wrongly named and we should probably rename it at some point. But to me it's pretty clear that this should not be part of the new `sale-percentage` plugin but rather stay in `shop-standard`.

I already added a filter on the sale label output created by `sale-percentage` so I will add a fixup here to use this filter and keep the feature working. :)

PR #102, opened 2020-05-27
Fixed wrong non-euro schema.org prices on single product template.

juanlopez4691:
### Ticket
- [Non-euro prices in the source code do not match the displayed prices](https://app.asana.com/0/809933051638353/1175777161118852/f)

### Description
-

### DIFF
--- a/src/Seo.php
+++ b/src/Seo.php
@@ -166,4 +166,22 @@ class Seo {
     return $data;
   }
 
+  /**
+   * Ensures the prices is the correct one, according to tax setings.
+   *
+   * @implements woocommerce_structured_data_product_offer
+   */
+  public static function adjustPrice($markup, $product) {
+    if (wc_prices_include_tax()) {
+      $product_price = wc_get_price_including_tax($product);
+    }
+    else {
+      $product_price = wc_get_price_excluding_tax($product);
+    }
+    $markup['price'] = $product_price;
+    $markup['priceSpecification']['price'] = $product_price;
+
+    return $markup;
+  }
+
 }
--- a/src/Seo.php
+++ b/src/Seo.php
@@ -167,7 +167,7 @@ class Seo {
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/Seo.php
```diff
@@ -166,4 +166,53 @@ public static function getProductBrand($data) {
     return $data;
   }
 
+  /**
+   * Fixes schema.org prices according to tax settings.
+   *
+   * When retrieving product prices to add into schema.org woocommerce is not
+   * considering if they already includes taxes, as set in the backend. This
+   * seems to be caused by aelya-currency-switcher currency conversion.
```
LucaPipolo, line 174:
```suggestion
   * seems to be caused by aelia-currency-switcher currency conversion.
```

src/Seo.php
```diff
@@ -166,4 +166,53 @@ public static function getProductBrand($data) {
...
+   *
+   * See https://github.com/woocommerce/woocommerce/blob/bfac625cdcda4d7d14e85ca3cf5534e86081e7bc/includes/class-wc-structured-data.php#L222-L223
```
LucaPipolo, line 176: Perhaps we could shorten this URL with Bitly or some similar tools?
At least this is what I do in such cases.

src/Seo.php
```diff
@@ -166,4 +166,53 @@ public static function getProductBrand($data) {
...
+   */
+  public static function adjustPrice($markup, $product) {
+    if (!class_exists('\WC_Aelia_CurrencyPrices_Manager')) {
+      return $markup;
+    }
+
+    $prices_include_tax = wc_prices_include_tax();
+
+    if ($product->get_type() === 'variable') {
+      $lowest  = $product->get_variation_price('min', $prices_include_tax);
```
LucaPipolo, line 188: Extra space.

```suggestion
      $lowest = $product->get_variation_price('min', $prices_include_tax);
```

PR #107, opened 2020-07-03
Fixed product availability entry in schema.org.

juanlopez4691:
### Ticket
- [Mismatched delivery time between front end an schema.org WOPA](https://app.asana.com/0/587433704548192/1181137040910250/f)

### Description
-

### DIFF
--- a/src/Seo.php
+++ b/src/Seo.php
@@ -253,7 +253,7 @@ class Seo {
    */
   public static function adjustAvailability($markup, $product) {
     if ($product->get_stock_quantity() || $product->backorders_allowed()) {
-      $markup['availability'] = 'http://schema.org/InStock';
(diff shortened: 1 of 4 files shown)


src/Seo.php
```diff
@@ -246,4 +246,16 @@ public static function adjustPrice($markup, $product) {
...
   }
 
+  /**
+   * Fixes schema.org product availability.
+   *
+   * @implements woocommerce_structured_data_product_offer
+   */
+  public static function adjustAvailability($markup, $product) {
+    if ($product->get_stock_quantity() || $product->backorders_allowed()) {
+      $markup['availability'] = 'http://schema.org/InStock';
```
LucaPipolo, line 256: `http://schema.org/InStock` is redirecting to `https://schema.org/InStock`.
Is not better to use the SSL protected version in our code?

PR #109, opened 2020-07-14
Fixed customer may be empty when using the WP CLI.

fabianmarz:
### Description
When using the WP CLI in combination with the customers base address to be `geolocation` the customer address will be empty causing not all hooks to work as expected.

In this specific case we wanted to generate product feeds but the tax calculation for some products failed as the base country could not be determined because it was set to geolocation which seems to not be available when using the CLI. As a fix we can set the customer to `base` which will use the base address of the shop instead.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -39,6 +39,10 @@ class Plugin {
    * @implements init
    */
   public static function preInit() {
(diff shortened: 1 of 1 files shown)


fabianmarz:
Closed due to obsolete.

PR #118, opened 2020-08-31
Added preload for main WooCommerce product image.

colourgarden:
### Ticket
- [Poor values for LCP on mobile devices | WOPA & LART](https://app.asana.com/0/809933051638353/1189713486865785/f)

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -402,6 +402,24 @@ class WooCommerce {
     }
   }
 
+  /*
+   * Adds preload tag for main product image to improve largest contentful paint.
+   *
+   * @implements wp_head
+   */
+  public static function preloadMainProductImage() {
+    global $post;
+
+    if (!$post || !$post->post_type === 'product') {
+      return;
+    }
+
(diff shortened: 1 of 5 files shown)


src/WooCommerce.php
```diff
@@ -398,6 +398,22 @@ public static function wp_head() {
...
+  public static function preloadMainProductImage() {
+    global $post;
+
+    if ($post && $post->post_type === 'product') {
+      $image = wp_get_attachment_image_src(get_post_thumbnail_id($post->ID), 'woocommerce_single');
+      if ($image) {
+        echo '<link rel="preload" href="' . $image[0] . '" as="image">';
+      }
+    }
+  }
```
juanlopez4691, line 415: You can get rid of the conditional nesting doing an early return.

```suggestion
  public static function preloadMainProductImage() {
    global $post;
    
    if (!$post || !$post->post_type === 'product') {
      return;
    }

    $image = wp_get_attachment_image_src(get_post_thumbnail_id($post->ID), 'woocommerce_single');
    if ($image) {
      echo '<link rel="preload" href="' . $image[0] . '" as="image">';
    }
  }
```

juanlopez4691 (approved):

PR #127, opened 2020-12-14
Changed product listing filter links into form buttons to avoid target pages from getting indexed.

Mauricio-Urrego:
### Ticket
- [Prevent links for filters that google can follow. | GACO](https://app.asana.com/0/30156186242982/1190953714418427/f)

### Description
-

### DIFF
--- a/src/ProductFilters/WidgetFilterDeliveryTime.php
+++ b/src/ProductFilters/WidgetFilterDeliveryTime.php
@@ -95,7 +95,7 @@ class WidgetFilterDeliveryTime extends \WC_Widget {
       }
       $chosen = in_array($delivery_time->term_id, $filter_values, TRUE);
       echo sprintf('<li class="%s">', $chosen ? $item_class . ' ' . $item_chosen_class : $item_class);
-      echo sprintf('<button type="submit" class="product-filter-term" formmethod="GET" name="delivery_time" value="%s">%s</button>', $value, $delivery_time->name);
+      echo sprintf('<a><button type="submit" class="product-filter-term" formmethod="GET" name="delivery_time" value="%s">%s</button></a></li>', $value, $delivery_time->name);
     }
     echo '</ul>';
     $this->widget_end($args);
--- a/src/ProductFilters/WidgetFilterDeliveryTime.php
+++ b/src/ProductFilters/WidgetFilterDeliveryTime.php
@@ -88,14 +88,11 @@ class WidgetFilterDeliveryTime extends \WC_Widget {
       else {
         $values = array_diff($values, [$delivery_time->term_id]);
       }
-      // Add filter values as query argument or remove parameter if empty.
-      // If the current value is selected the value will not be added.
-      if ($values) {
-        $value = implode(',', $values);
-      }
+      // Submit value as parameter.
+      $value = implode(',', $values);
       $chosen = in_array($delivery_time->term_id, $filter_values, TRUE);
       echo sprintf('<li class="%s">', $chosen ? $item_class . ' ' . $item_chosen_class : $item_class);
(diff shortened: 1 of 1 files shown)


src/ProductFilters/WidgetFilterDeliveryTime.php
```diff
@@ -92,13 +92,14 @@ public function widget($args, $instance) {
       // If the current value is selected the value will not be added.
       if ($values) {
         $link = add_query_arg('delivery_time', implode(',', $values));
```
sun, line 94:
```suggestion
```

src/ProductFilters/WidgetFilterDeliveryTime.php
```diff
@@ -92,13 +92,14 @@ public function widget($args, $instance) {
...
+        $value = implode(',', $values);
       }
       else {
         $link = remove_query_arg('delivery_time');
       }
```
sun, line 99:
```suggestion
```

src/ProductFilters/WidgetFilterDeliveryTime.php
```diff
@@ -92,13 +92,14 @@ public function widget($args, $instance) {
...
       $chosen = in_array($delivery_time->term_id, $filter_values, TRUE);
       echo sprintf('<li class="%s">', $chosen ? $item_class . ' ' . $item_chosen_class : $item_class);
-      echo sprintf('<a rel="nofollow" href="%s">%s</a></li>', $link, $delivery_time->name);
+      echo sprintf('<button type="submit" class="product-filter-term" formmethod="GET" name="delivery_time" value="%s" formnovalidate>%s</button>', $value, $delivery_time->name);
```
sun, line 102: The formnovalidate was just a suggestion, not sure whether it is needed if there is no surrounding form - probably not.
```suggestion
      echo sprintf('<button type="submit" class="product-filter-term" formmethod="GET" name="delivery_time" value="%s">%s</button>', $value, $delivery_time->name);
```

sun (changes requested):
$link appears to be obsolete. I also revised the PR title.

Mauricio-Urrego:
This has some to-do's. It needs to be tied to the checkbox that comes before the button. The button also needs to be able to remove the filter parameter when clicked a second time.

src/ProductFilters/WidgetFilterDeliveryTime.php
```diff
@@ -88,17 +88,11 @@ public function widget($args, $instance) {
...
-      }
-      else {
-        $link = remove_query_arg('delivery_time');
-      }
+      // Submit value as parameter.
+      $value = implode(',', $values);
       $chosen = in_array($delivery_time->term_id, $filter_values, TRUE);
       echo sprintf('<li class="%s">', $chosen ? $item_class . ' ' . $item_chosen_class : $item_class);
-      echo sprintf('<a rel="nofollow" href="%s">%s</a></li>', $link, $delivery_time->name);
+      echo sprintf('<form><a><button type="submit" class="product-filter-term" formmethod="GET" name="delivery_time" value="%s" style="border: none; background-color: #0000;">%s</button></a></form></li>', $value, $delivery_time->name);
```
fabianmarz, line 95: 1. Why do you use an a tag in conjunction with a button if the goal is to remove the links? 🤔 
2. This is very hard to read so please put some formatting in :)
3. Don't use inline styles. Use classes instead

Putting a form element around looks like the only way if we don't want to do it with JS. If we do so, why don't you put the action and method on the form element itself? Also what happens, if you want to select multiple filters?

sun: Can we output the wrapping form element only once around all filters?

PR #129, opened 2020-12-15
Added back in stock date to order items and the availability info.

fabianmarz:
- [Related Ticket](https://app.asana.com/0/30156186242982/1199570773208477/f)


src/WooCommerce.php
```diff
@@ -743,11 +749,11 @@ public static function woocommerce_display_item_meta($html, $item, $args) {
     $product_data_set = array_merge($data, static::getProductAttributes($product));
 
     // Display delivery time from woocommerce-german-market for each order item.
-    $delivery_time = wc_get_order_item_meta($item->get_id(), '_deliverytime');
-    if ($delivery_time && $delivery_time = get_term($delivery_time, 'product_delivery_times')) {
+    $delivery_time = \WGM_Template::get_deliverytime_string($product);
```
bogdanarizancu, line 752: It's all looking good, my only question is if it's ok to assume that all sites using `shop-standards` will automatically also use `woocommerce-german-market`. Maybe we should use what we previously had as fallback or something.

bogdanarizancu (approved):

PR #130, opened 2020-12-17
Fixed linter errors.

gyopiazza:
### Ticket
- [Styleboards](https://app.asana.com/0/1199152835480346/1199172788290607)

### Description
The related ticket only brought me to discover a few bugs in the linting configuration of the shop-standards plugin, which are fixed in this PR. The relevant changes are:

- Added dependencies: "stylelint": "^13.8.0", "stylelint-config-standard": "^20.0.0”
- Added stylelint config: .stylelintrc.json

### Note
- I've added the global variables as arguments of the IIFE in admin.js, instead of marking them as exceptions (like it's done in main.js).
- Gulp generated a new untracked file: dist/shopstandards.js


.stylelintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "stylelint-config-standard"
+}
```
fabianmarz, line 3: Seems your editor is configured wrongly. Please ensure there's a new line at the end of each file.

assets/scripts/main.js
```diff
@@ -86,7 +86,7 @@
   $(document).ready(() => {
     // Disable copy/paste actions on billing email fields.
     if (shop_standards_settings.emailConfirmationEmail === 'yes') {
-      $('#billing_email, #billing_email_confirmation').each(function() {
+      $('#billing_email, #billing_email_confirmation').each(function x() {
```
fabianmarz, line 89: Why did you name this function?

assets/scripts/main.js
```diff
@@ -86,7 +86,7 @@
   $(document).ready(() => {
     // Disable copy/paste actions on billing email fields.
     if (shop_standards_settings.emailConfirmationEmail === 'yes') {
```
fabianmarz, line 88: Wny didn't you pass in the parameter `shop_standards_settings` as well, if you did it for the `admin.js` file?

dist/scripts/shopstandards.js
```diff
@@ -0,0 +1,138 @@
+"use strict";
```
fabianmarz, line 1: I think this file is generated due to a wrong build process. Not sure what the correct command is, or if there's an npm script in the `package.json`. AFAIK this file is generated if the file is concatenated. So this could possibly prevented by `$ gulp build --concat=false` and/or `$ gulp build --concat=false --production`.

fabianmarz (changes requested):

assets/scripts/main.js
gyopiazza: I only fixed all the "errors" that the linter was showing, but you're totally right.

assets/scripts/main.js
gyopiazza: The linter was requiring it, I'll double check.

gyopiazza: Confirmed: it was needed by the linter as a "warning". I've now disabled the "func-names" check in `.eslintrc` so we don't get the warning.

dist/scripts/shopstandards.js
gyopiazza: I've run both and the `--production` flag is needed to avoid generating `shopstandards.js`. I've also added "build" npm script to make it more obvious `npm run build`

.stylelintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "stylelint-config-standard"
```
wrux, line 2: @gyopiazza I created a netz Stylelint ruleset recently, you can see it used in NUTR-LOPRO: https://github.com/netzstrategen/nutr-lopro/blob/master/.stylelintrc.json

package.json
```diff
@@ -2,6 +2,9 @@
   "version": "1.37.2",
   "title": "shop-standards",
   "description": "Standard refinements for e-commerce websites.",
+  "scripts": {
+    "build": "gulp build --concat=false --production"
```
wrux, line 6: @gyopiazza This is based on a very old version of gulp-task-collection. In version 3.0 we added some improvements so it might make sense to upgrade it here. 

If you upgrade gulp-task-collection then you can probably change these build scripts up a bit

```suggestion
    "watch": "gulp",
    "build": "gulp build:production"
```

fabianmarz:
Closed due to fixes in #143.

PR #131, opened 2020-12-18
Fixed length postcode validation for Liechtenstein.

bogdanarizancu:
### Ticket
- [Valid four-digit Liechtenstein postcodes marked as invalid, stopping orders](https://app.asana.com/0/587433704548192/1199387326781845/f)

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -909,7 +909,7 @@ class WooCommerce {
         break;
 
       case 'LI':
-        $valid = (bool) preg_match('@^((948[5-9]{1})|(949[0-7]{1}))$@', $postcode);
+        $valid = (bool) preg_match('@^((948[5-9]{1})|(949[0-8]{1}))$@', $postcode);
         break;
 
       case 'NL':
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -909,7 +909,7 @@ class WooCommerce {
         break;
 
       case 'LI':
-        $valid = (bool) preg_match('@^((948[5-9]{1})|(949[0-8]{1}))$@', $postcode);
+        $valid = (bool) preg_match('@^((948[5-9])|(949[0-8]))$@', $postcode);
         break;
 
       case 'NL':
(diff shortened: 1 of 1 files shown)


juanlopez4691 (approved):

src/WooCommerce.php
```diff
@@ -909,7 +909,7 @@ public static function woocommerce_validate_postcode($valid, $postcode, $country
         break;
 
       case 'LI':
-        $valid = (bool) preg_match('@^(948[5-9])|(949[0-7])$@', $postcode);
+        $valid = (bool) preg_match('@^((948[5-9]{1})|(949[0-7]{1}))$@', $postcode);
```
sun, line 912: `{1}` is the default of each character.
```suggestion
        $valid = (bool) preg_match('@^((948[5-9])|(949[0-7]))$@', $postcode);
```

sun (changes requested):

src/WooCommerce.php
bogdanarizancu: It is added to limit to 1 character length, it otherwise allows 5 digit postcodes, I have already tested this.

sun: I do not see a repeater instruction in the expression, so this does not match 5 digits.

See for yourself here: https://regex101.com/r/jIZPN7/1

src/WooCommerce.php
```diff
@@ -909,7 +909,7 @@ public static function woocommerce_validate_postcode($valid, $postcode, $country
         break;
 
       case 'LI':
-        $valid = (bool) preg_match('@^(948[5-9])|(949[0-7])$@', $postcode);
+        $valid = (bool) preg_match('@^((948[5-9]{1})|(949[0-8]{1}))$@', $postcode);
```
sun, line 912:
```suggestion
        $valid = (bool) preg_match('@^((948[5-9])|(949[0-8]))$@', $postcode);
```

sun (approved):
As discussed above

src/WooCommerce.php
bogdanarizancu: Now I see what you mean, thank you.

PR #132, opened 2021-01-04
 Added feature to disable related products on a per-product basis.

wrux:
### Ticket
- [Hide "related products" section in frontend for certain products.](https://app.asana.com/0/30156186242982/1199229224946101/f)

### Description
-

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -1182,6 +1182,7 @@ class WooCommerce {
     $product_id = $product->get_id();
     if (get_post_meta($product->get_id(), '_' . Plugin::PREFIX . '_disable_related_products', true) === 'yes') {
       add_filter('woocommerce_product_related_posts_force_display', '__return_false', 40, 2);
+      add_filter('woocommerce_related_products', '__return_empty_array');
     }
   }
 
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -1181,7 +1181,7 @@ class WooCommerce {
     global $product;
     $product_id = $product->get_id();
     if (get_post_meta($product->get_id(), '_' . Plugin::PREFIX . '_disable_related_products', true) === 'yes') {
-      add_filter('woocommerce_product_related_posts_force_display', '__return_false', 99999, 2);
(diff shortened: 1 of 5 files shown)


Mauricio-Urrego (approved):

src/WooCommerce.php
```diff
@@ -1158,4 +1174,15 @@ public static function getVariationIdByAttributes($product_id, array $attributes
...
   }
 
+  /**
+   * Disables output of related products if over-ride checkbox is enabled.
+   */
+  public static function disableRelatedProducts(): void {
+    global $product;
+    $product_id = $product->get_id();
+    if (get_post_meta($product->get_id(), '_' . Plugin::PREFIX . '_disable_related_products', true) === 'yes') {
+      add_filter('woocommerce_product_related_posts_force_display', false, $product_id, 20, 2);
```
wrux, line 1184: If you look at the docs for [add_filter](https://developer.wordpress.org/reference/functions/add_filter/), it shows that the second argument needs to be a callable and third is priority. 

The second argument ` '__return_false'` is equivalent to `function() { return false; }` 

Perhaps this is working because $product_id is a really high number, so it's adding the filter so it executes last.

Do you think this will work?
```suggestion
      add_filter('woocommerce_product_related_posts_force_display',  '__return_false', 99999, 20, 2);
```

Mauricio-Urrego: Oh... I played around with priority as well by giving it a higher priority (e.g. priority 10). I did not consider it would need lower priority.

Mauricio-Urrego: Does it just ignore the last parameter? And wouldn't this make accepted args: 20? I'm confused.

wrux: Oh yea my suggestion wasn't the best example. Would this work? 

```
add_filter('woocommerce_product_related_posts_force_display',  '__return_false', 99999, 2);
```

Mauricio-Urrego: Yes this works too so it's definitely the priority. Just need to put this on staging.

PR #136, opened 2021-03-11
Removed trailing slashes in product filter to avoid unnecessary redirects.

Mauricio-Urrego:
### Ticket
- [Product category filter: Attribute facet filter submits to a wrong URL causing redirect (adding 0.5s)](https://app.asana.com/0/30156186242982/1200027730245172/f)

### Description
-

### DIFF
--- a/src/ProductFilters/WidgetLayeredNav.php
+++ b/src/ProductFilters/WidgetLayeredNav.php
@@ -31,10 +31,16 @@ class WidgetLayeredNav extends \WC_Widget_Layered_Nav {
   }
 
   /**
-   * @todo Implement output.
+   * Remove trailing slash for performance increase (avoids redirects).
    */
   protected function layered_nav_dropdown($terms, $taxonomy, $query_type) {
-    return parent::layered_nav_dropdown($terms, $taxonomy, $query_type);
+    ob_start();
+    $found = parent::layered_nav_dropdown($terms, $taxonomy, $query_type);
+    $output = ob_get_clean();
+
+    $output =  preg_replace("/gartenmoebel\//", "gartenmoebel", $output);;
+    echo $output;
+    return $found;
   }
 
 }
--- a/src/ProductFilters/WidgetLayeredNav.php
+++ b/src/ProductFilters/WidgetLayeredNav.php
@@ -38,7 +38,7 @@ class WidgetLayeredNav extends \WC_Widget_Layered_Nav {
     $found = parent::layered_nav_dropdown($terms, $taxonomy, $query_type);
     $output = ob_get_clean();
 
-    $output =  preg_replace("/gartenmoebel\//", "gartenmoebel", $output);;
+    $output =  preg_replace("/gartenmoebel\//", "gartenmoebel", $output);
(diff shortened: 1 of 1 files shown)


sun (changes requested):
We can't do this in shop-standards, because a shop does not necessarily have permalinks configured to omit trailing slashes.

However, GACO has permalinks configured without trailing slash, so the layered_nav_dropdown should produce the correct links already. Can you check whether they might get overridden by custom code or another plugin (Yoast?)?

src/ProductFilters/WidgetLayeredNav.php
```diff
@@ -31,10 +31,16 @@ protected function layered_nav_list($terms, $taxonomy, $query_type) {
...
-   * @todo Implement output.
+   * Remove trailing slash for performance increase (avoids redirects).
    */
   protected function layered_nav_dropdown($terms, $taxonomy, $query_type) {
-    return parent::layered_nav_dropdown($terms, $taxonomy, $query_type);
+    ob_start();
+    $found = parent::layered_nav_dropdown($terms, $taxonomy, $query_type);
+    $output = ob_get_clean();
+
+    $output =  preg_replace("/gartenmoebel\//", "gartenmoebel", $output);;
```
bogdanarizancu, line 41:
```suggestion
    $output =  preg_replace("/gartenmoebel\//", "gartenmoebel", $output);
```

bogdanarizancu (changes requested):

src/ProductFilters/WidgetLayeredNav.php
Mauricio-Urrego: Oops

Mauricio-Urrego:
> We can't do this in shop-standards, because a shop does not necessarily have permalinks configured to omit trailing slashes.
> 
> However, GACO has permalinks configured without trailing slash, so the layered_nav_dropdown should produce the correct links already. Can you check whether they might get overridden by custom code or another plugin (Yoast?)?

We decided on a call to go with a different strategy here. Will try removing trailing slashes using home_url filter.

Mauricio-Urrego:
Went with this filter now as the strategy. Filter is being called in too late though. Needs further testing...

wp-content/plugins/custom/src/WooCommerce.php:
 ```
/**
   * Prevents unnecessary redirect due to '?/' for get parameters.
   *
   * @implements home_url
   */
  public static function home_url($url, $path, $orig_scheme, $blog_id) {
    if ($path && strpos($url, '/?') !== FALSE) {
      $url = str_replace('/?', '?', $url);
    }
    return $url;
  }
```
wp-content/plugins/custom/src/Plugin.php:
```
// Remove trailing slash form the filter form action.
add_filter('home_url', __NAMESPACE__ . '\WooCommerce::home_url', 1, 4);
```

Mauricio-Urrego:
This PR has been replaced with: https://github.com/netzstrategen/gaco/pull/621

Mauricio-Urrego:
This PR has been replaced with: https://github.com/netzstrategen/gaco/pull/621

PR #139, opened 2021-05-12
Fixed first page link in pagination causes redirect to actual initial listing page.

Mauricio-Urrego:
### Ticket
- [Prevent pagination links to page 1 that contain the page suffix. ](https://app.asana.com/0/507013885678070/1200321868255267/f)

### Description
-

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -209,7 +209,7 @@ class Plugin {
      */
     add_filter('paginate_links', function ($link) {
       $link = preg_replace('@/page/1/?(\?|$)@', '$1', $link);
-      return $link;
+      return user_trailingslashit($link);
     });
   }
 
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -203,6 +203,14 @@ class Plugin {
 
     // Disables output of related products if over-ride checkbox is enabled.
     add_action('woocommerce_after_single_product_summary', __NAMESPACE__ . '\WooCommerce::disableRelatedProducts', 0, 2);
+
+    /**
+     * Removes the `/page/1` suffix from archive URLs.
+     */
+    add_filter('paginate_links', function ($link) {
+      $link = preg_replace('@/page/1/?(\?|$)@', '$1', $link);
+      return $link;
+    });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


bogdanarizancu (approved):

src/Plugin.php
```diff
@@ -203,6 +203,14 @@ public static function init() {
 
     // Disables output of related products if over-ride checkbox is enabled.
     add_action('woocommerce_after_single_product_summary', __NAMESPACE__ . '\WooCommerce::disableRelatedProducts', 0, 2);
+
+    /**
+     * Removes the `/page/1` suffix from archive URLs.
+     */
```
sun, line 209: Only a inline comment (single line summary) here; PHPDoc block is only for the separate function definition.

src/Plugin.php
```diff
@@ -203,6 +203,14 @@ public static function init() {
...
+    add_filter('paginate_links', function ($link) {
+      $link = preg_replace('@/page/1/?(\?|$)@', '$1', $link);
+      return user_trailingslashit($link);
+    });
```
sun, line 213: Please move to a separate static function/callback.

sun (changes requested):

src/Plugin.php
```diff
@@ -203,6 +203,9 @@ public static function init() {
 
     // Disables output of related products if over-ride checkbox is enabled.
     add_action('woocommerce_after_single_product_summary', __NAMESPACE__ . '\WooCommerce::disableRelatedProducts', 0, 2);
+
+    // Removes the `/page/1` suffix from archive URLs.
```
sun, line 207: Avoid backticks in PHP code whenever possible, they are the shorthand operator for executing shell commands:
https://www.php.net/manual/en/language.operators.execution.php

```suggestion
    // Removes the suffix '/page/1' from archive URLs.
```

src/Plugin.php
```diff
@@ -346,4 +349,14 @@ public static function loadTextdomain() {
     load_plugin_textdomain(static::L10N, FALSE, static::L10N . '/languages/');
   }
 
+  /**
+   * Removes the `/page/1` suffix from archive URLs.
```
sun, line 353:
```suggestion
   * Removes the suffix '/page/1' from archive URLs.
```

sun (approved):
@Mauricio-Urrego I corrected the PR title. It should state the original problem that is being addressed by the proposed change.

Please squash-merge using the corrected PR title as commit message.

PR #140, opened 2021-05-12
Patches pagination in Elementor product widget.

Mauricio-Urrego:
### Ticket
- [Elementor: pagination clicks reload full page - no anchor](https://app.asana.com/0/30156186242982/1200264058586621/f)

### Description
- Relevant issue: https://github.com/elementor/elementor/issues/4703

Patched by adding an id tag to the widget that page links can jump to.


assets/scripts/main.js
```diff
@@ -110,4 +110,14 @@
       $('.wcppec-checkout-buttons').hide();
     }
   });
+  
+  // Fixes missing anchor on elementor widget pagination.
+  // Must be combined with <a name="widget"></a> in title.
+  // https://github.com/elementor/elementor/issues/4703
+  $(document).ready(() => {
+    $('.elementor-widget-woocommerce-products').attr('id', $('.elementor-widget-woocommerce-products').attr('data-id'));
```
sun, line 118:
```suggestion
    const $widget = $('.elementor-widget-woocommerce-products');
    if (!$widget.attr('id')) {
      $widget.attr('id', $widget.attr('data-id'));
    }
```
You can use the same `$widget` as context for the selector looking for all links below.

assets/scripts/main.js
```diff
@@ -110,4 +110,14 @@
...
+    $('.elementor-widget-woocommerce-products .page-numbers a').each((i, a) => {
+      $(a).attr('href', $(a).attr('href') + '#' + $('.elementor-widget-woocommerce-products').attr('id'));
```
sun, line 120: The first argument (which you named `i`) should contain the full jQuery object of the link already, no?

In addition, there may be multiple product listing widgets on a single page, so you should find the closest to the link:
```suggestion
      $(a).attr('href', $(a).attr('href') + '#' + $(a).closest('.elementor-widget-woocommerce-products').attr('id'));
```

sun (changes requested):

sun:
@Mauricio-Urrego I tested and my last suggestions on this PR should work correctly. Please have another look.

image

sun (approved):

sun:
The code should work, seems like something else is conflicting.

fabianmarz (changes requested):
Blocking for now as the compiled file is missing.

fabianmarz:
Just a quick heads up. This is not working because some other elementor function (`handleAnchorLinks`)[https://github.com/elementor/elementor/blob/2cd3d50d8d87edde9cd3c820f5a0f8b1f2c4f24f/assets/dev/js/frontend/utils/anchors.js#L64] code is hooking into all anchors and puts a `e.preventDefault` if it's an internal link 🤔 

I'm currently looking into a viable solution.

Mauricio-Urrego:
> Just a quick heads up. This is not working because some other elementor function (`handleAnchorLinks`)[https://github.com/elementor/elementor/blob/2cd3d50d8d87edde9cd3c820f5a0f8b1f2c4f24f/assets/dev/js/frontend/utils/anchors.js#L64] code is hooking into all anchors and puts a `e.preventDefault` if it's an internal link 🤔
> 
> @Mauricio-Urrego, could you please test my commit? I remove the click event listeners to prevent elementor from treating these links as anchor target links. I guess it's save to remove the click listeners all together as there should not be any attached.

Works for me! Just need to commit compiled assets then?

fabianmarz (approved):

PR #142, opened 2021-05-19
Fixed back-in-stock not showing for variable products in products listing.

gyopiazza:
### Ticket
- [Variable Products: missing "back in stock"-date (overview pages)](https://app.asana.com/0/809933051638353/1200306165872411)

### Description
The "back in stock" date was visible for "simple products" only. The reason is that the logic was only getting the date from the "master product", which only works for "simple products" since such date is stored on each product variation. This PR extends the current logic by gathering the "EarliestBackInStock" date either from the master product (in the case of simple products) or from variations (in the case of variable products).

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -177,7 +177,7 @@ class WooCommerce {
       }
     }
 
-    if ($product->backorders_allowed() && $back_in_stock_date = self::getBackInStockMeta($product)) {
+    if ($product->backorders_allowed() && $back_in_stock_date = self::getEarliestBackInStock($product)) {
       if ($date_string = static::getBackInStockDateString($back_in_stock_date)) {
         $stock['availability'] = '<strong>' . sprintf(__('Back in stock %s', Plugin::L10N), $date_string) . '</strong>';
       }
@@ -987,18 +987,26 @@ class WooCommerce {
 
   /**
    * Retrieves the earliest "back in stock" date from a product or its variations.
+   * If there's stock, the product is considered available and nothing is returned.
    * @param WC_Product
    *    The product object.
    * @return string
    *    The earliest back in stock date, if any.
    */
-  public static function getBackInStockMeta($product) {
+  public static function getEarliestBackInStock($product) {
+    if (!$product->get_stock_quantity()) {
+      return '';
+    }
+
     $meta_key = '_' . Plugin::PREFIX . '_back_in_stock_date';
     $back_in_stock_date = get_post_meta($product->get_id(), $meta_key, TRUE);
 
     if ($product->is_type('variable')) {
       $variations = $product->get_available_variations();
       foreach ($variations as $variation) {
+        if ($variation['max_qty'] > 0) {
+          return '';
+        }
         $variation_stock_date = get_post_meta($variation['variation_id'], $meta_key, TRUE);
         if (!$back_in_stock_date || $back_in_stock_date > $variation_stock_date) {
           $back_in_stock_date = $variation_stock_date;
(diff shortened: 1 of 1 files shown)


src/WooCommerce.php
```diff
@@ -985,6 +985,41 @@ public static function getTaxonomyTermsAsSelectOptions($taxonomy) {
     return is_wp_error($terms) ? [] : $term_options;
   }
 
+  /**
+   * Retrieves the earliest "back in stock" date from a product or its variations.
+   * If there's stock, the product is considered available and nothing is returned.
+   * @param WC_Product
+   *    The product object.
+   * @return string
+   *    The earliest back in stock date, if any.
```
fabianmarz, line 994:
```suggestion
   * If there's stock, the product is considered available and nothing is returned.
   *
   * @param WC_Product
   *    The product object.
   *
   * @return string
   *    The earliest back in stock date, if any.
```

src/WooCommerce.php
```diff
@@ -985,6 +985,41 @@ public static function getTaxonomyTermsAsSelectOptions($taxonomy) {
...
+   */
+  public static function getEarliestBackInStock($product) {
```
fabianmarz, line 996:
```suggestion
  public static function getEarliestBackInStock(\WC_Product $product): string {
```

src/WooCommerce.php
```diff
@@ -985,6 +985,41 @@ public static function getTaxonomyTermsAsSelectOptions($taxonomy) {
...
+      return '';
+    }
+
+    $meta_key = '_' . Plugin::PREFIX . '_back_in_stock_date';
+    $back_in_stock_date = get_post_meta($product->get_id(), $meta_key, TRUE) ?? '';
+
+    if ($product->is_type('variable')) {
+      $variations = $product->get_available_variations();
+      foreach ($variations as $variation) {
+        $variation_obj = wc_get_product($variation['variation_id']);
```
fabianmarz, line 1007: Maybe a more suitable name? 🤔 
```suggestion
        $product_variation = wc_get_product($variation['variation_id']);
```

src/WooCommerce.php
```diff
@@ -985,6 +985,41 @@ public static function getTaxonomyTermsAsSelectOptions($taxonomy) {
...
+        $variation_stock_date = get_post_meta($variation['variation_id'], $meta_key, TRUE);
+        // If a variation has stock or doesn't have a "back in stock" date
+        // it's considered to be immediately available for purchase.
+        if ($variation_obj->get_stock_quantity() > 0 || !$variation_stock_date) {
```
fabianmarz, line 1011: Do we also need to consider `backorder_allowed`?

src/WooCommerce.php
gyopiazza: I'll change it, but I named it that way to differenciate it between the "array like" returned by `get_available_variations` as opposed to the Object returned by `wc_get_product`.

src/WooCommerce.php
gyopiazza: I don't think we do, since the check was already implemented before calling the `getEarliestBackInStock` function:

https://github.com/netzstrategen/wordpress-shop-standards/pull/142/files#diff-bd92fcb7a0c2355394fe0edb44db3c7d0648b4c4d382c92b18dab73cc4865b35R180

https://github.com/netzstrategen/wordpress-shop-standards/pull/142/files#diff-bd92fcb7a0c2355394fe0edb44db3c7d0648b4c4d382c92b18dab73cc4865b35R180

fabianmarz (approved):
To me the code looks fine. If it's tested by a PM you can proceed merging it :) Nice job!

PR #145, opened 2021-05-27
Added optional title for product attribute listing pages.

Mauricio-Urrego:
### Ticket
- [Product Attribute Listings: Optional field for custom page title instead of attribute/term/filter name](https://app.asana.com/0/809933051638353/1200232830814947/f)

### Description
-

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.38.4",
+  "version": "1.38.5",
   "title": "shop-standards",
   "description": "Standard refinements for e-commerce websites.",
   "dependencies": {
--- a/plugin.php
+++ b/plugin.php
@@ -40,6 +40,7 @@ register_uninstall_hook(__FILE__, __NAMESPACE__ . '\Schema::uninstall');
 add_action('plugins_loaded', __NAMESPACE__ . '\Plugin::loadTextdomain');
 add_action('init', __NAMESPACE__ . '\Plugin::preInit', 0);
 add_action('init', __NAMESPACE__ . '\Plugin::init', 20);
+add_action('init', __NAMESPACE__ . '\ProductAttributePageTitle::init', 20);
 add_filter('woocommerce_get_settings_pages', __NAMESPACE__ . '\Settings::woocommerce_get_settings_pages');
 add_action('admin_init', __NAMESPACE__ . '\Admin::init');
 
--- a/plugin.php
+++ b/plugin.php
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Shop Standards
-  Version: 1.38.4
+  Version: 1.38.5
   Text Domain: shop-standards
   Description: Standard refinements for e-commerce websites.
   Author: netzstrategen
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -290,14 +290,14 @@ class Plugin {
     // Add alternative title to page categorization.
     register_field_group([
     	'key' => 'group_60afa34a62839',
-    	'title' => 'Other',
+    	'title' => __('Other', Plugin::L10N),
     	'fields' => [
     		[
     			'key' => 'field_60afa47226f5f',
-    			'label' => 'Page Name',
+    			'label' => __('Page Name', Plugin::L10N),
     			'name' => 'page_name',
     			'type' => 'text',
-    			'instructions' => 'Only changes how the name appears on the page. Leaves the filter drop-down menu alone.',
+    			'instructions' => __('Only changes how the name appears on the page. Leaves the filter drop-down menu alone.', Plugin::L10N),
     			'required' => 0,
     			'conditional_logic' => 0,
     			'wrapper' => [
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -81,7 +81,6 @@ class Plugin {
         'menu_title' => __('Hide "Add to Cart" button', Plugin::L10N),
         'parent_slug' => 'woocommerce',
       ]);
-      ProductAttributePageTitle::init();
       Plugin::register_acf();
     }
 
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -74,6 +74,11 @@ class Plugin {
     }
 
     if (function_exists('register_field_group')) {
+      // ACF
+      add_filter( 'acf/location/rule_types', __NAMESPACE__ . '\WooCommerce::rule_types');
+      add_filter( 'acf/location/rule_values/wc_prod_attr', __NAMESPACE__ . '\WooCommerce::wc_prod_attr');
+      add_filter( 'acf/location/rule_match/wc_prod_attr', __NAMESPACE__ . '\WooCommerce::wc_prod_attr_2', 10, 3);
+
       acf_add_options_sub_page([
         'page_title' => __('Hide "Add to Cart" button', Plugin::L10N),
         'menu_title' => __('Hide "Add to Cart" button', Plugin::L10N),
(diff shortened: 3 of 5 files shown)


sun (dismissed):
1. The proposed code is not in line with the coding standards and coding style at all. Did you copy it from somewhere else? 😕 
2. Please ensure to set a proper PR title - it's not clear at all that this is about listing pages of product attributes.
3. Each feature in this plugin should be located in a dedicated class. Can you move all of the code into a new ProductAttributePageTitle class, please?
4. Please double-check whether you really need to load all product attribute taxonomies to register the field for all of them in the register_field_group() call. I'm positive there must be other possible values for `'operator'` that require the parameter to exist without testing its value.  I also don't really understand what the filters for the rule are doing and why they are necessary.
5. Did you consider to implement this without ACF? It might require less code.
6. If you cannot avoid a separate group in the backend UI, then at least rename the group from "Other" to "Listing Page".  Please also replace the randomly generated hash in the keys with an appropriate name.
7. "page_name" is a proper concept in WordPress already (the post "slug"), using it for something else here would be very confusing. Please rename it to e.g. "page_title".
8. I expect this code to also replace the value of the page title on the page when a custom value has been set in the field.  This should also apply to the title in the HTML head. Therefore, it would probably be best to conditionally adjust the term name in the global $queried_object variable through a filter before the page rendering starts.

Mauricio-Urrego:
@sun You're absolutely right. I just implemented something that doesn't use ACF and it works perfectly fine. Will push soon.

src/ProductAttributePageTitle.php
```diff
@@ -0,0 +1,74 @@
...
+          'wrapper' => [
+            'width' => '',
+            'class' => '',
+            'id' => '',
+          ],
+          'default_value' => '',
+          'placeholder' => '',
+          'prepend' => '',
+          'append' => '',
+          'maxlength' => '',
```
fabianmarz, line 42: I think there is lots of stuff which can be dropped as it's the default. E.g. all of this code + probably more. 
```suggestion
```

fabianmarz (changes requested):

src/ProductAttributePageTitle.php
```diff
@@ -0,0 +1,74 @@
...
+        ],
+      ],
+      'location' => [
+        [
+          [
+            'param'    => 'taxonomy',
+            'operator' => '==',
+            'value'    => 'all',
```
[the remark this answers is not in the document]
Mauricio-Urrego: Hmm, so taxonomy == product maybe was it?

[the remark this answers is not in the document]
Mauricio-Urrego: @gyopiazza Do you know how to narrow this down more?

src/Plugin.php
```diff
@@ -79,6 +81,7 @@ public static function init() {
         'menu_title' => __('Hide "Add to Cart" button', Plugin::L10N),
         'parent_slug' => 'woocommerce',
       ]);
+      ProductAttributePageTitle::init();
```
wrux, line 84: I think you can just add a WP hook to invoke this, similar to how Plugin::init() is called here: https://github.com/netzstrategen/wordpress-shop-standards/blob/2c9771e76c06076b917fcd08353c70310649dbc0/plugin.php#L42

Perhaps on the next line you can just insert the following

```php
add_action('init', __NAMESPACE__ . '\ProductAttributePageTitle::init', 20);
```

src/ProductAttributePageTitle.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Netzstrategen\ShopStandards;
+
+/**
+ * Allows optional customization of page title.
+ */
+class ProductAttributePageTitle {
+
+  public static function init() {
```
wrux, line 10: This is introducing a dependency of ACF. It's installed on most websites nowadays, however I think this plugin might be installed on some sites without ACF.

wrux (changes requested):
@Mauricio-Urrego I've found a tutorial for adding a field to the tax form [here](https://rudrastyh.com/wordpress/add-custom-fields-to-taxonomy-terms.html). I think it's  better to go with a solution that does not use ACF as some websites with this plugin may not have ACF installed.

src/Plugin.php
Mauricio-Urrego: @wrux These are wordpress tags not woocommerce attribute/term settings fields.

src/ProductAttributePageTitle.php
Mauricio-Urrego: @wrux Shop standards is already using ACF.

Mauricio-Urrego:
> @Mauricio-Urrego I've found a tutorial for adding a field to the tax form [here](https://rudrastyh.com/wordpress/add-custom-fields-to-taxonomy-terms.html). I think it's better to go with a solution that does not use ACF as some websites with this plugin may not have ACF installed.

@wrux These are wordpress tags not woocommerce attribute/term settings fields.

src/ProductAttributePageTitle.php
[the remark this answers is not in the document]
Mauricio-Urrego: @wrux Either way, it would be good if there was a suggestion added here on this topic.

sun:
@Mauricio-Urrego You should be able to follow a tutorial for adding fields to taxonomy terms, as the product attribute terms are regular taxonomy terms under the hood. Here are some additional pointers:

- https://wordpress.stackexchange.com/questions/371083/add-custom-field-to-woocommerce-add-new-attribute-edit-page
- https://pippinsplugins.com/adding-custom-meta-fields-to-taxonomies/

ACF:
- https://wordpress.stackexchange.com/questions/302352/is-it-possible-to-add-custom-fields-to-a-woocommerce-attribute-term
- https://www.hereswhatidid.com/2016/08/add-acf-fields-product-attributes-woocommerce/

Mauricio-Urrego:
> @Mauricio-Urrego You should be able to follow a tutorial for adding fields to taxonomy terms, as the product attribute terms are regular taxonomy terms under the hood. Here are some additional pointers:
> 
>     * https://wordpress.stackexchange.com/questions/371083/add-custom-field-to-woocommerce-add-new-attribute-edit-page
> 
>     * https://pippinsplugins.com/adding-custom-meta-fields-to-taxonomies/
> 
> 
> ACF:
> 
>     * https://wordpress.stackexchange.com/questions/302352/is-it-possible-to-add-custom-fields-to-a-woocommerce-attribute-term
> 
>     * https://www.hereswhatidid.com/2016/08/add-acf-fields-product-attributes-woocommerce/

Thanks Sunny  - I found the same links while implementing this but their solutions bring me to this: https://github.com/netzstrategen/wordpress-shop-standards/pull/145#discussion_r642506075

Mauricio-Urrego:
@sun I have re-implemented your suggested solution from the last link shared in your comment [here](https://github.com/netzstrategen/wordpress-shop-standards/pull/145#issuecomment-860612055).

src/ProductAttributePageTitle.php
Mauricio-Urrego: [Here](https://github.com/netzstrategen/wordpress-shop-standards/blob/master/acf/brands.hide-to-cart-button-product-categories-brands.json)

wrux (approved):

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.38.4",
+  "version": "1.38.5",
```
sun, line 2: A new feature should increase the minor version, not the patch version. See https://semver.org/

plugin.php
```diff
@@ -40,6 +40,7 @@ function classloader($class) {
 add_action('plugins_loaded', __NAMESPACE__ . '\Plugin::loadTextdomain');
 add_action('init', __NAMESPACE__ . '\Plugin::preInit', 0);
 add_action('init', __NAMESPACE__ . '\Plugin::init', 20);
+add_action('init', __NAMESPACE__ . '\ProductAttributePageTitle::init', 20);
```
sun, line 43: The contents of the new init function should have been added directly to Plugin::init() instead.  The point of our structure is to avoid loading unnecessary code on pages that do not need it.  By registering ProductAttributePageTitle::init() for the init hook here, the whole class is loaded unconditionally on all pages.

src/Plugin.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Netzstrategen\ShopStandards;
 
+use Netzstrategen\ShopStandards\ProductAttributePageTitle;
```
sun, line 5: Unused import. This should have been noticed and reverted before merging.

src/ProductAttributePageTitle.php
```diff
@@ -0,0 +1,67 @@
...
+    foreach( $attribute_terms as $attribute_term ) {
+      $group_filter[] = [[
+        'param'    => 'taxonomy',
+        'operator' => '==',
+        'value'    => $attribute_term,
+      ]];
+    }
+
+    // Register ACF field.
+    ProductAttributePageTitle::register_acf_page_title($group_filter);
```
sun, line 23: Every frontend page view loads all attribute taxonomies now, even though the frontend may not need the data.

Maybe you can move this block into a new method `admin_init()` in this class and only invoke it (from elsewhere) when `is_admin()` is true?

src/ProductAttributePageTitle.php
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Allows optional customization of page title.
+ */
+class ProductAttributePageTitle {
+
+  public static function init() {
+    $attribute_terms = wc_get_attribute_taxonomy_names();
+    $group_filter = [];
+
+    foreach( $attribute_terms as $attribute_term ) {
```
sun, line 14: There are still various coding style issues in the merged code.

src/ProductAttributePageTitle.php
```diff
@@ -0,0 +1,67 @@
...
+      ],
+      'location' => $group_filter,
+      'menu_order' => 0,
+      'position' => 'normal',
+      'style' => 'default',
+      'label_placement' => 'top',
+      'instruction_placement' => 'label',
+      'hide_on_screen' => '',
+      'active' => true,
+      'description' => '',
```
sun, line 55: Many of these are default values that can be omitted.
```suggestion
        ],
      ],
      'location' => $group_filter,
```

sun (commented):
@Mauricio-Urrego Could you create a follow-up PR to address these issues, please?

PR #146, opened 2021-06-03
Fixes trailing slash issue on pagination.

Mauricio-Urrego:
### Ticket
- [Prevent pagination links to page 1 that contain the page suffix. ](https://app.asana.com/0/0/1200322193267937/f)

### Description
- slash before /page in regex is not needed.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -355,8 +355,7 @@ class Plugin {
    * @return string
    */
   public static function paginate_links($link) {
-    $link = preg_replace('@/page/1/?(\?|$)@', '$1', $link);
-    return user_trailingslashit($link);
+    return preg_replace('@page/1/?(\?|$)@', '$1', $link);
   }
 
 }


src/Plugin.php
```diff
@@ -355,8 +355,7 @@ public static function loadTextdomain() {
    * @return string
    */
   public static function paginate_links($link) {
-    $link = preg_replace('@/page/1/?(\?|$)@', '$1', $link);
-    return user_trailingslashit($link);
+    return preg_replace('@page/1/?(\?|$)@', '$1', $link);
```
fabianmarz, line 358: @Mauricio-Urrego, why are u reverting the changes you introduced in https://github.com/netzstrategen/wordpress-shop-standards/pull/139/files already?

Mauricio-Urrego: There is an issue with trailingslashit adding an encoded trailing slash to query parameters whenever pagination is used and a filter is active. They accumulate on every page navigation like '?filter_marken=vitra%2F%2F%2F'. I will find a better solution though. This one does not work on all sites which was the original intention of trailingslashit.

PR #147, opened 2021-06-09
Fixed fatal error when getting stock quantity for product variations.

fabianmarz:
### Description
- This PR fixes a possible fatal error as the the product variation is an array instead of an object and therefore the called method is not available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.38.3",
+  "version": "1.38.4",
   "title": "shop-standards",
   "description": "Standard refinements for e-commerce websites.",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


sun:
Did you actually encounter the fatal error?

IIRC, the "object" parameter was introduced recently. Maybe the site you encountered this on still has an older version of WooCommerce?

fabianmarz:
Yeah we encountered it on FECO which seems to run on 4.2.2. But indeed, that's why I used the `'object'` argument (and could not remember). Thanks for catching!

So maybe it makes more sense to close the PR and instead ensure the other project is updated. What do you think?

fabianmarz:
Can be closed as this is relevant for an outdated WooCommerce version only. Will be automatically fixed once the related project is updated.

PR #150, opened 2021-06-18
Fixed Elementor JavaScript runs before scroll position fix.

Mauricio-Urrego:
### Ticket
- [Elementor: pagination clicks reload full page - no anchor](https://app.asana.com/0/30156186242982/1200264058586621/f)

### Description
- Solves a race condition where jquery did not add the necessary tag fast enough for the browser to navigate to it.


assets/scripts/main.js
```diff
@@ -132,7 +132,7 @@
 
   // Fixes missing anchor on elementor widget pagination.
   // https://github.com/elementor/elementor/issues/4703
-  $(document).ready(() => {
+  document.addEventListener('DOMContentLoaded', () => {
```
sun, line 135: What this change really means is that this block of the JS should be moved into a separate file and that separate file should be loaded with a higher priority on the page, so that it is loaded before the JS file of Elementor.

In addition, WordPress supports dependencies between scripts. Using them would be even more appropriate.  However, this is a reversed dependency and that is not supported easily. Therefore, loading it with a higher priority is the simplest solution.

sun (changes requested):

assets/scripts/main.js
Mauricio-Urrego: Can I change it back to jquery and then make the priority 999 for the action?
`add_action('wp_enqueue_scripts', __CLASS__ . '::wp_enqueue_scripts', 999);`

src/Plugin.php
```diff
@@ -301,6 +302,12 @@ public static function wp_enqueue_scripts() {
     ]);
   }
 
+  public static function wp_enqueue_elementor_script() {
+    $git_version = static::getGitVersion();
+    $suffix = defined('SCRIPT_DEBUG') && SCRIPT_DEBUG ? '' : '.min';
+    wp_enqueue_script(Plugin::PREFIX, static::getBaseUrl() . '/dist/scripts/vendor/elementor' . $suffix . '.js', ['jquery'], $git_version, TRUE);
```
Mauricio-Urrego, line 308: @sun Is this right? It does not seem to be loading the script?

assets/scripts/vendor/elementor.js
```diff
@@ -0,0 +1,17 @@
+/* global jQuery */
+
+(function pageLoad($) {
```
sun, line 3:
```suggestion
(function ($) {
```

src/Plugin.php
sun, line 308: The first parameter is the internal identifier of the script, which allows other scripts to use it as a dependency (in the same way you're specifying "jquery" here). The identifier must be unique for each script; otherwise WP assumes that the script has been loaded already.
```suggestion
    wp_enqueue_script(Plugin::PREFIX . '/elementor', static::getBaseUrl() . '/dist/scripts/vendor/elementor' . $suffix . '.js', ['jquery'], $git_version, TRUE);
```

src/Plugin.php
```diff
@@ -188,6 +188,7 @@ public static function init() {
 
     // Enqueues plugin scripts and styles.
     add_action('wp_enqueue_scripts', __CLASS__ . '::wp_enqueue_scripts');
+    add_action('wp_enqueue_scripts', __CLASS__ . '::wp_enqueue_elementor_script', 999);
```
sun, line 191: The priority needs to be less than the default (10) in order to be triggered before Elementor:
```suggestion
    add_action('wp_enqueue_scripts', __CLASS__ . '::wp_enqueue_elementor_script', 9);
```

Note that Elementor also appears to invoke an own hook when enqueueing its own scripts, so you could try this:
```suggestion
    add_action('elementor/frontend/before_enqueue_scripts', __CLASS__ . '::wp_enqueue_elementor_script');
```

sun (changes requested):

assets/scripts/vendor/elementor.js
Mauricio-Urrego: Screen Shot 2021-06-21 at 18 28 04

Mauricio-Urrego: I named it different now than the main.js function though.

src/Plugin.php
Mauricio-Urrego: So I think something is wrong with the compiling still because this only works without the `$suffix`

src/Plugin.php
Mauricio-Urrego: Yes I got the direction confused, I have made it 9 now.

Mauricio-Urrego: Very cool, I did not know there was a hook for this. Thank you, that makes it a lot more reliable.

src/Plugin.php
sun: Okay, please remove `$suffix` for now.  The `.min.js` file is only compiled if `gulp` is invoked with the `--production` flag.  Most likely the `npm run build` command was not updated for this plugin yet.  Your branch only contains the unminified file, too.  This is very minor, the script very small.  We can improve the frontend performance later.

Mauricio-Urrego: Ok removed.

src/Plugin.php
```diff
@@ -301,6 +302,11 @@ public static function wp_enqueue_scripts() {
     ]);
   }
 
+  public static function wp_enqueue_elementor_script() {
+    $git_version = static::getGitVersion();
+    wp_enqueue_script(Plugin::PREFIX . '/elementor', static::getBaseUrl() . '/dist/scripts/vendor/elementor' . '.js', ['jquery'], $git_version);
```
sun, line 307:
```suggestion
    wp_enqueue_script(Plugin::PREFIX . '/elementor', static::getBaseUrl() . '/dist/scripts/vendor/elementor.js', ['jquery'], $git_version);
```

sun (approved):

PR #151, opened 2021-06-21
Improved performance of optional title for product attribute listing pages.

Mauricio-Urrego:
Fixups for https://github.com/netzstrategen/wordpress-shop-standards/pull/145

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.40.1",
+  "version": "1.40.4",
   "title": "shop-standards",
   "description": "Standard refinements for e-commerce websites.",
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.41.0",
+  "version": "1.40.1",
   "title": "shop-standards",
   "description": "Standard refinements for e-commerce websites.",
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.40.3",
+  "version": "1.41.0",
   "title": "shop-standards",
   "description": "Standard refinements for e-commerce websites.",
   "dependencies": {
--- a/plugin.php
+++ b/plugin.php
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Shop Standards
-  Version: 1.41.0
+  Version: 1.40.1
   Text Domain: shop-standards
   Description: Standard refinements for e-commerce websites.
   Author: netzstrategen
--- a/plugin.php
+++ b/plugin.php
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Shop Standards
-  Version: 1.40.3
+  Version: 1.41.0
   Text Domain: shop-standards
   Description: Standard refinements for e-commerce websites.
   Author: netzstrategen
--- a/plugin.php
+++ b/plugin.php
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Shop Standards
-  Version: 1.40.1
+  Version: 1.40.4
   Text Domain: shop-standards
   Description: Standard refinements for e-commerce websites.
   Author: netzstrategen
--- a/plugin.php
+++ b/plugin.php
@@ -40,7 +40,6 @@ register_uninstall_hook(__FILE__, __NAMESPACE__ . '\Schema::uninstall');
 add_action('plugins_loaded', __NAMESPACE__ . '\Plugin::loadTextdomain');
 add_action('init', __NAMESPACE__ . '\Plugin::preInit', 0);
 add_action('init', __NAMESPACE__ . '\Plugin::init', 20);
-add_action('init', __NAMESPACE__ . '\ProductAttributePageTitle::init', 20);
 add_filter('woocommerce_get_settings_pages', __NAMESPACE__ . '\Settings::woocommerce_get_settings_pages');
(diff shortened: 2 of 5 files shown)


src/Plugin.php
```diff
@@ -109,6 +107,11 @@ public static function init() {
     PlusProducts::init();
 
     if (is_admin()) {
+      if (function_exists('register_field_group')) {
+        // Adds optional customization of page title.
+        ProductAttributePageTitle::admin_init();
+      }
```
sun, line 113: Maybe we could move this to `Admin::init()`?

src/Plugin.php
```diff
@@ -208,6 +211,9 @@ public static function init() {
 
     // Removes the suffix '/page/1' from archive URLs.
     add_filter('paginate_links', __CLASS__  . '::paginate_links');
+
+    // Front-end output of custom page title.
+    add_filter('woocommerce_page_title', __CLASS__.'::woocommerce_page_title');
```
sun, line 216: Did you try wrapping this into `is_term()`?

src/ProductAttributePageTitle.php
```diff
@@ -7,23 +7,22 @@
...
-        'param'    => 'taxonomy',
-        'operator' => '==',
-        'value'    => $attribute_term,
-      ]];
+    foreach ($attribute_terms as $attribute_term) {
+      $group_filter[] = [
+        [
+          'param' => 'taxonomy',
+          'operator' => '==',
+          'value' => $attribute_term,
```
sun, line 19: Are you sure that you cannot use a LIKE operator here to avoid loading all taxonomies?
```suggestion
          'operator' => 'LIKE',
          'value' => 'pa_%',
```

src/ProductAttributePageTitle.php
```diff
@@ -40,28 +39,9 @@ public static function register_acf_page_title($group_filter) {
...
-      'position' => 'normal',
-      'style' => 'default',
-      'label_placement' => 'top',
-      'instruction_placement' => 'label',
-      'hide_on_screen' => '',
-      'active' => true,
-      'description' => '',
     ]);
   }
-
```
sun, line 58: There should be a blank line before and after a class method.

sun (changes requested):
Thanks!  Can you also revisit the PR title, please?

src/Plugin.php
Mauricio-Urrego: is_term is deprecated and term_exists leads to a false. Not sure if I should use the optional params?

src/ProductAttributePageTitle.php
Mauricio-Urrego: I think I tried this but I will try again to be sure.

src/Plugin.php
Mauricio-Urrego: Yes, good idea, it looks strange here.

src/ProductAttributePageTitle.php
Mauricio-Urrego: Ok, yes I am sure.

src/Admin.php
```diff
@@ -41,6 +41,11 @@ public static function init() {
     // Removes admin notices about updating to version and connecting to WooCommerce.com.
     add_filter('woocommerce_helper_suppress_admin_notices', '__return_true');
     add_filter('woocommerce_helper_suppress_connect_notice', '__return_true');
+
+    if (function_exists('register_field_group')) {
+      // Adds optional customization of page title.
```
sun, line 46:
```suggestion
      // Add field to set custom page title for product attribute listing pages.
```

src/Plugin.php
```diff
@@ -208,6 +206,9 @@ public static function init() {
 
     // Removes the suffix '/page/1' from archive URLs.
     add_filter('paginate_links', __CLASS__  . '::paginate_links');
+
+    // Front-end output of custom page title.
```
sun, line 210:
```suggestion
    // Override page title on product attribute pages.
```

src/Plugin.php
sun: 1. The callback should remain in the feature class, so that it is clear to which feature this hook registration belongs.
2. There should be a space before and after the string concatenation operator.
```suggestion
    add_filter('woocommerce_page_title', __NAMESPACE__ . '\ProductAttributePageTitle::woocommerce_page_title');
```

src/Plugin.php
```diff
@@ -364,4 +365,12 @@ public static function paginate_links($link) {
     return $link;
   }
 
+  /**
+   * Front-end output of custom page title.
```
sun, line 369:
```suggestion
   * Overrides page title on product attribute pages.
```

sun (changes requested):

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.40.0",
+  "version": "1.41.0",
```
sun, line 2: This is a performance fix and doesn't introduce new functionality, so only the patch version should increase.
```suggestion
  "version": "1.40.1",
```

plugin.php
```diff
@@ -2,7 +2,7 @@
 
 /*
   Plugin Name: Shop Standards
-  Version: 1.40.0
+  Version: 1.41.0
```
sun, line 5:
```suggestion
  Version: 1.40.1
```

sun (approved):
Looks much better now 👍 

Please ensure that the functionality still works (both in the backend and in the frontend) before merging this PR.

PR #152, opened 2021-06-21
Added anchors to woocommerce product filters.

Mauricio-Urrego:
### Ticket
- [Shop: set scroll-position to the beginning of the filter after setting a filter](https://app.asana.com/0/587433704548192/1181771954608505/f)

### Description
- Need to combine this with already open PR [here](https://github.com/netzstrategen/wopa/pull/806). Needs to either be in shop-standards completely or in theme completely.


src/ProductFilters/WidgetLayeredNav.php
```diff
@@ -34,7 +34,15 @@ protected function layered_nav_list($terms, $taxonomy, $query_type) {
    * @todo Implement output.
    */
   protected function layered_nav_dropdown($terms, $taxonomy, $query_type) {
-    return parent::layered_nav_dropdown($terms, $taxonomy, $query_type);
+    ob_start();
+    $found = parent::layered_nav_dropdown($terms, $taxonomy, $query_type);
```
sun, line 38: The parent class WC_Widget_Layered_Nav invokes several hooks before and after building the filter links and forms, which should allow you to adjust them without manipulating already rendered HTML.

sun (changes requested):

src/ProductFilters/WidgetLayeredNav.php
Mauricio-Urrego: No hooks available. Can do this upstream with woocommerce?

Mauricio-Urrego:
Replaced by https://github.com/netzstrategen/wopa/pull/806

PR #155, opened 2021-07-14
Fixed fatal errors if advanced-custom-fields plugin is not installed.

Mauricio-Urrego:
### Ticket
- []()

### Description
-

### DIFF
--- a/src/ProductAttributePageTitle.php
+++ b/src/ProductAttributePageTitle.php
@@ -20,7 +20,9 @@ class ProductAttributePageTitle {
     }
 
     // Register ACF field.
-    ProductAttributePageTitle::register_acf_page_title($group_filter);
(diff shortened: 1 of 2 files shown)


src/ProductAttributePageTitle.php
```diff
@@ -20,7 +20,9 @@ public static function init() {
     }
 
     // Register ACF field.
-    ProductAttributePageTitle::register_acf_page_title($group_filter);
+    if (function_exists('register_field_group')) {
```
bogdanarizancu, line 23: @Mauricio-Urrego I think this would make more sense if we added it before it was actually invoked, like on line 33. This is only a suggestion, feel free to decide yourself, nice catch.

bogdanarizancu (approved):

PR #158, opened 2021-07-20
Fixed delivery time in order confirmation use value at order time

joram77:
### Ticket
- [Customer account: Change/delete delivery time in order confirmation](https://app.asana.com/0/30156186242982/1200426661151439/f)

### Description
-Changes have been made to staging to show the delivery time in the order confirmation as it was specified at the moment of the order. This instead of using the current product delivery time property


Mauricio-Urrego (approved):
Code looks good to me - I did not test but it seems to be QA'd and work as expected.

PR #165, opened 2021-09-27
Added cronjob to remove outdated back-in-stock metadata for products without Moeve ID.

bogdanarizancu:
### Ticket
- [shop-standards: Obsolete back in stock dates are not removed from database](https://app.asana.com/0/30156186242982/1200496606141091/f)

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -38,7 +38,7 @@ class Plugin {
    *
    * @var string
    */
-  const CRON_EVENT_ENSURE_BACK_IN_STOCK = Plugin::PREFIX . '/cron/ensure-back-in-stock';
+  const CRON_EVENT_ENSURE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
 
   /**
    * Plugin initialization method with the lowest possible priority.
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -182,6 +182,11 @@ class Plugin {
       add_filter('woocommerce_de_avoid_check_same_delivery_time_show_parent', '__return_true');
     }
 
+    add_action(self::CRON_EVENT_ENSURE_BACK_IN_STOCK, __NAMESPACE__ . '\WooCommerce::cron_ensure_back_in_stock');
+    if (!wp_next_scheduled(static::CRON_EVENT_ENSURE_BACK_IN_STOCK)) {
+      wp_schedule_event(strtotime('3am tomorrow'), 'daily', self::CRON_EVENT_ENSURE_BACK_IN_STOCK);
+    }
+
     // Prefetches DNS entries for particular resources.
     add_filter('wp_resource_hints', __NAMESPACE__ . '\Performance::wp_resource_hints', 10, 2);
     // Preloads scripts and loads styles asynchronously.
@@ -217,11 +222,6 @@ class Plugin {
 
     // Override page title on product attribute pages.
     add_filter('woocommerce_page_title', __NAMESPACE__ . '\ProductAttributePageTitle::woocommerce_page_title');
-
-    add_action(self::CRON_EVENT_ENSURE_BACK_IN_STOCK, __NAMESPACE__ . '\Schema::cron_ensure_back_in_stock');
-    if (!wp_next_scheduled(static::CRON_EVENT_ENSURE_BACK_IN_STOCK)) {
-      wp_schedule_event(strtotime('3am tomorrow'), 'daily', self::CRON_EVENT_ENSURE_BACK_IN_STOCK);
-    }
   }
 
   /**
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -34,6 +34,13 @@ class Plugin {
   public static $version = '';
 
   /**
+   * Cron event name for checking database indexes.
+   *
+   * @var string
+   */
+  const CRON_EVENT_ENSURE_BACK_IN_STOCK = Plugin::PREFIX . '/cron/ensure-back-in-stock';
+
+  /**
    * Plugin initialization method with the lowest possible priority.
    *
    * @implements init
@@ -210,6 +217,11 @@ class Plugin {
 
     // Override page title on product attribute pages.
     add_filter('woocommerce_page_title', __NAMESPACE__ . '\ProductAttributePageTitle::woocommerce_page_title');
+
+    add_action(self::CRON_EVENT_ENSURE_BACK_IN_STOCK, __NAMESPACE__ . '\Schema::cron_ensure_back_in_stock');
+    if (!wp_next_scheduled(static::CRON_EVENT_ENSURE_BACK_IN_STOCK)) {
+      wp_schedule_event(strtotime('3am tomorrow'), 'daily', self::CRON_EVENT_ENSURE_BACK_IN_STOCK);
+    }
   }
 
   /**
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -38,7 +38,7 @@ class Plugin {
    *
    * @var string
    */
-  const CRON_EVENT_ENSURE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
+  const CRON_EVENT_REMOVE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
 
   /**
    * Plugin initialization method with the lowest possible priority.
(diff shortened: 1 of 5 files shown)


src/Plugin.php
```diff
@@ -33,6 +33,13 @@ class Plugin {
    */
   public static $version = '';
 
+  /**
+   * Cron event name for checking database indexes.
```
sun, line 37:
```suggestion
   * Cron event name for removing outdated back-in-stock product metadata.
```

src/Plugin.php
```diff
@@ -33,6 +33,13 @@ class Plugin {
...
+   *
+   * @var string
+   */
+  const CRON_EVENT_ENSURE_BACK_IN_STOCK = Plugin::PREFIX . '/cron/ensure-back-in-stock';
```
sun, line 41:
```suggestion
  const CRON_EVENT_ENSURE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
```

src/Plugin.php
```diff
@@ -210,6 +217,11 @@ public static function init() {
 
     // Override page title on product attribute pages.
     add_filter('woocommerce_page_title', __NAMESPACE__ . '\ProductAttributePageTitle::woocommerce_page_title');
+
+    add_action(self::CRON_EVENT_ENSURE_BACK_IN_STOCK, __NAMESPACE__ . '\Schema::cron_ensure_back_in_stock');
+    if (!wp_next_scheduled(static::CRON_EVENT_ENSURE_BACK_IN_STOCK)) {
+      wp_schedule_event(strtotime('3am tomorrow'), 'daily', self::CRON_EVENT_ENSURE_BACK_IN_STOCK);
+    }
```
sun, line 224: 1. Let's move this up to after line 171 so that it is grouped with the main back-in-stock functionality, please.

1. Please also move the callback from Schema to WooCommerce, below the existing callback for the back-in-stock functionality.  (Ideally we'd group them into an own ProductBackInStock class, but that's for a future issue.)

src/Schema.php
```diff
@@ -25,4 +25,28 @@ public static function deactivate() {
   public static function uninstall() {
   }
 
+  /**
+   * Cron event callback to ensure proper database indexes.
```
sun, line 29:
```suggestion
   * Cron event callback to remove outdated back-in-stock product metadata.
```

src/Schema.php
```diff
@@ -25,4 +25,28 @@ public static function deactivate() {
...
+   */
+  public static function cron_ensure_back_in_stock() {
+    global $wpdb;
+    $ids = $wpdb->get_col($wpdb->prepare("SELECT ID FROM {$wpdb->posts}
+      INNER JOIN {$wpdb->postmeta} pm ON {$wpdb->posts}.ID = pm.post_id
+      LEFT JOIN {$wpdb->postmeta} moeve ON {$wpdb->posts}.ID = moeve.post_id AND moeve.meta_key LIKE %s
+      WHERE pm.meta_key = %s
+        AND pm.meta_value <= CURRENT_DATE()
```
sun, line 37: The current date should be injected as a query variable from PHP (`date('Y-m-d')`), because the date in MySQL should not be trusted.
```suggestion
        AND pm.meta_value <= %s
```

src/Schema.php
```diff
@@ -25,4 +25,28 @@ public static function deactivate() {
...
+        AND moeve.meta_id IS NULL", [
+          '_woocommerce-moeve_id_%',
```
sun, line 39: Did you test and confirm that this works correctly?  To my knowledge, `wpdb::prepare()` treats percent signs specially as they are also the variable placeholders.  But maybe it works here because it appears at the end of string?

src/Schema.php
```diff
@@ -25,4 +25,28 @@ public static function deactivate() {
...
+          '_' . Plugin::PREFIX . '_back_in_stock_date',
+    ]));
+
+    foreach ($ids as $id) {
+      delete_post_meta($id, '_' . Plugin::PREFIX . '_back_in_stock_date');
+      if (function_exists('wc_get_product')) {
+        $product = wc_get_product($product);
+        $product->save();
+      }
```
sun, line 48: 1. Undefined variable `$product`.
1. I would have expected to only see the WooCommerce API here…? https://woocommerce.github.io/code-reference/classes/WC-Data.html#method_delete_meta_data
```suggestion
      if ($product = wc_get_product($id)) {
        $product->delete_meta_data('_' . Plugin::PREFIX . '_back_in_stock_date');
        $product->save();
      }
```

sun (changes requested):
Thank you!  I left a few comments that we should adjust before proceeding with this.

src/Schema.php
bogdanarizancu: If this runs on a project were we do not have WooCommerce available it will fail without the `function_exists` check, won't it?

src/Plugin.php
bogdanarizancu: I moved after the entire check for WooCommerce German Market, since it should not be dependent on that.

src/Schema.php
bogdanarizancu: @sun I did test and the result is correct, when I had it inside the sql string like:
```mysql
LEFT JOIN {$wpdb->postmeta} moeve ON {$wpdb->posts}.ID = moeve.post_id AND moeve.meta_key LIKE '_woocommerce-moeve_id_%'
```
then it would not work, but prepared like this it will return the same amount of results and the full raw query.

src/Schema.php
sun: Good point, but then I would wrap the whole callback and perhaps even the cron event registration altogether into that condition.  Here in the middle of the loop / at the end of the action is a bit late. 😉

src/WooCommerce.php
```diff
@@ -186,6 +186,29 @@ public static function woocommerce_get_availability($stock, $product) {
...
+    global $wpdb;
+    $ids = $wpdb->get_col($wpdb->prepare("SELECT ID FROM {$wpdb->posts}
+      INNER JOIN {$wpdb->postmeta} pm ON {$wpdb->posts}.ID = pm.post_id
+      LEFT JOIN {$wpdb->postmeta} moeve ON {$wpdb->posts}.ID = moeve.post_id AND moeve.meta_key LIKE %s
+      WHERE pm.meta_key = %s
+        AND pm.meta_value <= %s
+        AND moeve.meta_id IS NULL", [
+          '_woocommerce-moeve_id_%',
+          '_' . Plugin::PREFIX . '_back_in_stock_date',
+          date('Y-m-d H:i:s'),
```
sun, line 202: "Any time today" should be considered obsolete, so we should omit the time.
```suggestion
          date('Y-m-d'),
```

src/Plugin.php
```diff
@@ -175,6 +182,11 @@ public static function init() {
       add_filter('woocommerce_de_avoid_check_same_delivery_time_show_parent', '__return_true');
     }
 
+    add_action(self::CRON_EVENT_ENSURE_BACK_IN_STOCK, __NAMESPACE__ . '\WooCommerce::cron_ensure_back_in_stock');
+    if (!wp_next_scheduled(static::CRON_EVENT_ENSURE_BACK_IN_STOCK)) {
+      wp_schedule_event(strtotime('3am tomorrow'), 'daily', self::CRON_EVENT_ENSURE_BACK_IN_STOCK);
+    }
```
sun, line 188: 1. If WooCommerce is not installed, we can skip the entire cron event.
2. Also update the cron event name constant accordingly to "remove" instead of "ensure".
3. Ditto for the callback function name.
4. The initial timestamp should be 3am today; otherwise you have to wait for one day until you see results and can confirm that it works.
```suggestion
    if (function_exists('wc_get_product')) {
      add_action(self::CRON_EVENT_REMOVE_BACK_IN_STOCK, __NAMESPACE__ . '\WooCommerce::cron_remove_back_in_stock');
      if (!wp_next_scheduled(static::CRON_EVENT_REMOVE_BACK_IN_STOCK)) {
        wp_schedule_event(strtotime('03:00:00'), 'daily', self::CRON_EVENT_REMOVE_BACK_IN_STOCK);
      }
    }
```

src/Plugin.php
```diff
@@ -33,6 +33,13 @@ class Plugin {
    */
   public static $version = '';
 
+  /**
+   * Cron event name for removing outdated back-in-stock product metadata.
+   *
+   * @var string
+   */
+  const CRON_EVENT_ENSURE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
```
sun, line 41:
```suggestion
  const CRON_EVENT_REMOVE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
```

src/WooCommerce.php
```diff
@@ -186,6 +186,29 @@ public static function woocommerce_get_availability($stock, $product) {
...
+    ]));
+    foreach ($ids as $id) {
+      if (function_exists('wc_get_product') && $product = wc_get_product($id)) {
```
sun, line 205:
```suggestion
      if ($product = wc_get_product($id)) {
```

src/WooCommerce.php
```diff
@@ -186,6 +186,29 @@ public static function woocommerce_get_availability($stock, $product) {
...
+   * Cron event callback to remove outdated back-in-stock product metadata.
+   */
+  public static function cron_ensure_back_in_stock() {
+    global $wpdb;
+    $ids = $wpdb->get_col($wpdb->prepare("SELECT ID FROM {$wpdb->posts}
+      INNER JOIN {$wpdb->postmeta} pm ON {$wpdb->posts}.ID = pm.post_id
+      LEFT JOIN {$wpdb->postmeta} moeve ON {$wpdb->posts}.ID = moeve.post_id AND moeve.meta_key LIKE %s
+      WHERE pm.meta_key = %s
+        AND pm.meta_value <= %s
+        AND moeve.meta_id IS NULL", [
```
sun, line 199: I'd still recommend to use a table alias for the primary table and the back in stock date meta field, too.  It makes the whole query more readable.
```suggestion
    $ids = $wpdb->get_col($wpdb->prepare("SELECT p.ID FROM {$wpdb->posts} p
      INNER JOIN {$wpdb->postmeta} backinstock ON p.ID = backinstock.post_id
      LEFT JOIN {$wpdb->postmeta} moeve ON p.ID = moeve.post_id AND moeve.meta_key LIKE %s
      WHERE backinstock.meta_key = %s
        AND backinstock.meta_value <= %s
        AND moeve.meta_id IS NULL", [
```

sun (approved):
Thanks!  This looks good now - only a few things could be polished a little.

PR #168, opened 2021-11-09
Fixed performance bottleneck caused by earliest back in stock functionality.

chrisbaltazar:
### Ticket
https://app.asana.com/0/784106262441860/1201312032738062/f

### Description
The problem was the amount of times the function woocommerce_de_get_deliverytime_string_label_string was been called in the listing pages, triggering even more subqueries from it, at least ny removing the Product Variation objects from the calling stack I was able to reduce the local page loading by more than half... 

I refactored also the function to format the back in stock date which was doing an extra condition to make sure the date was a valid future date when the it has been done in the previous step in all cases.

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -188,7 +188,7 @@ class WooCommerce {
     }
 
     if ($product->backorders_allowed() && $back_in_stock_date = self::getEarliestBackInStock($product)) {
-      $date_string = static::getFormatBackInStockDateString($back_in_stock_date);
+      $date_string = static::getFormattedBackInStockDateString($back_in_stock_date);
       $stock['availability'] = '<strong>' . sprintf(__('Back in stock %s', Plugin::L10N), $date_string) . '</strong>';
     }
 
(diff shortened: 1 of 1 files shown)


src/WooCommerce.php
```diff
@@ -1032,20 +1030,22 @@ public static function getEarliestBackInStock(\WC_Product $product): string {
       return '';
     }
 
-    $meta_key = self::FIELD_BACK_IN_STOCK_DATE;
-    $back_in_stock_date = get_post_meta($product->get_id(), $meta_key, TRUE) ?? '';
+    $back_in_stock_date = get_post_meta($product->get_id(), self::FIELD_BACK_IN_STOCK_DATE, TRUE) ?? '';
```
bogdanarizancu, line 1033:
```suggestion
    $back_in_stock_date = get_post_meta($product->get_id(), self::FIELD_BACK_IN_STOCK_DATE, TRUE);
```

src/WooCommerce.php
```diff
@@ -1063,12 +1063,14 @@ public static function getEarliestBackInStock(\WC_Product $product): string {
    * @implements woocommerce_de_get_deliverytime_string_label_string
    */
   public static function woocommerce_de_get_deliverytime_string_label_string($label_string, $product) {
-    if (!$product) {
+    // Avoid the execution of this function for product variants which can be
+    // evaluated over the parent product in the next method
+    if (!$product || $product instanceof \WC_Product_Variation) {
```
bogdanarizancu, line 1068:
```suggestion
    if (!$product) {
```
This change might interfere with other contexts where it shouldn't.

bogdanarizancu (approved):

bogdanarizancu (approved):
LGTM

PR #169, opened 2021-11-10
Added WC filter to display inline error messages to checkout

chrisbaltazar:
### Ticket
https://app.asana.com/0/587433704548192/1181759168801807/f

### Description
Added WC filter to display inline error messages to checkout


src/WooCommerceCheckout.php
```diff
@@ -28,6 +28,36 @@ public static function init() {
       add_filter('woocommerce_checkout_fields', __CLASS__ . '::addConfirmationEmailCheckoutField');
       add_action('woocommerce_checkout_process', __CLASS__ . '::checkConfirmationEmailField');
     }
+
+    add_filter( 'woocommerce_form_field', __CLASS__ . '::woocommerceFormField', 10, 4 );
```
fabianmarz, line 32: For the hook we usually add a small comment what it's used for. This you don't have to jump to the function to see if it's the right one. (Some hooks might miss the comments but we should get more consistent with it)

```suggestion

    // Add checkout error messages.
    add_filter( 'woocommerce_form_field', __CLASS__ . '::woocommerceFormField', 10, 4 );
```

src/WooCommerceCheckout.php
```diff
@@ -28,6 +28,36 @@ public static function init() {
...
+   * @return string
+   */
+  public static function woocommerceFormField($field, $key, $args, $value) {
+    if (strpos($field, '</label>') === FALSE) {
+      return $field;
+    }
+
+    $error = [];
+    if ($args['required'] ?? FALSE) {
+      $errorMessage = sprintf(__('%s is a required field.', 'woocommerce'), $args['label']);
```
fabianmarz, line 49: Local variables should be in lower_snake_case while class properties are in lowerCamelCase.
```suggestion
      $error_message = sprintf(__('%s is a required field.', 'woocommerce'), $args['label']);
```

src/WooCommerceCheckout.php
```diff
@@ -28,6 +28,36 @@ public static function init() {
...
+      $error[]      = '<span class="error-required" style="display:none">' . $errorMessage . '</span>';
```
fabianmarz, line 50: Variables should not be aligned. Imagine bigger multi line arrays holding keys sorted by name. If you add another, longer key, it's possible you will have to adjust dozens of lines as based on this single addition. 
```suggestion
      $error[] = '<span class="error-required" style="display:none">' . $error_message . '</span>';
```

src/WooCommerceCheckout.php
```diff
@@ -28,6 +28,36 @@ public static function init() {
...
+    }
+    if ($args['validate'] ?? FALSE) {
+      $errorMessage = __('Invalid field', 'woocommerce');
+      $error[]      = '<span class="error-validate" style="display:none">' . $errorMessage . '</span>';
+    }
+
+    if (empty($error)) {
+      return $field;
+    }
+    return substr_replace($field, join('', $error), strpos($field, '</label>'), 0);
```
fabianmarz, line 60: Does it make sense to check if there is an error to result in a single return value?
```suggestion
    if ($error) {
      $field = substr_replace($field, join('', $error), strpos($field, '</label>'), 0);
    }
    return $field;
    
```

fabianmarz (changes requested):
can't comment on the file itself, but why are there so many deletions in the `package-lock.json` file?

src/WooCommerceCheckout.php
chrisbaltazar: OK I get that, but in this case you mean put 2 different comments in general? like one short description in the hook and then the whole thing in the function?

src/WooCommerceCheckout.php
chrisbaltazar: this variable differentiation is also a drupal convention? or why is it like this?

src/WooCommerceCheckout.php
chrisbaltazar: I guess is just a matter of taste here, but pretty much same thing

chrisbaltazar:
> can't comment on the file itself, but why are there so many deletions in the `package-lock.json` file?

is just the result of the synchronisation after doing a simple npm install 'cause we were having some dependency issues, so nothing actually changed by us

chrisbaltazar:
HI @fabianmarz thanks for your comments, I replied all of them but I really think we need to sit and check how we handle the code standards after all, I just followed the guide in notion for this again but I don't get the same format you mention, so I think is worth to clarify for this and further cases :)

src/WooCommerceCheckout.php
fabianmarz: Yes https://www.drupal.org/docs/develop/standards/coding-standards#s-functions-and-variables we agreed on using the different cases through the files. But I agree we should rather stick to a single convention instead of mixing it depending on the type. E.g. use lowerCamelCase. I think this needs to be discussed mainly with @sun.

fabianmarz (approved):
Code LGTM (apart from the massive package lock removal). But maybe it wasn't committed in a longer time.

src/WooCommerceCheckout.php
chrisbaltazar: Alright, worth to mention thanks

PR #170, opened 2021-11-16
Fixed error messages displayed in checkout

chrisbaltazar:
### Ticket
https://app.asana.com/0/587433704548192/1201339113105126

### Description
Display one error message at the time instead of all of them at once over each field


geist-ahnen (approved):
Looks good to me.

PR #171, opened 2021-11-22
Added feature to include the main category in product links.

chrisbaltazar:
### Ticket
https://app.asana.com/0/784106262441860/1201093170276968/f

### Description
Added function to enforce the main category on product links by using a custom option in order to activate this behavior when needed


src/Plugin.php
```diff
@@ -112,6 +112,7 @@ public static function init() {
     WooCommerceSalutation::init();
     WooCommerceCheckout::init();
     PlusProducts::init();
+    Products::init();
```
bogdanarizancu, line 115: Since this PHP class is doing something very specific, I think this is a too general class name, so something similar to the existing `ProductAttributePageTitle`, maybe like `ProductPermalink`, might be a good idea.

src/Products.php
```diff
@@ -0,0 +1,136 @@
+<?php
+
+namespace Netzstrategen\ShopStandards;
+
+/**
+ * Meant to handle special Product features.
```
bogdanarizancu, line 6: This is a very general description and does not create much value, I think it explain exactly why it is trying to accomplish, it could be something in the line of
```suggestion
 * Allows to optionally include the main category in the permalink for SEO.
```

src/Products.php
```diff
@@ -0,0 +1,136 @@
...
+   *
+   * @var string
+   */
+  private string $product_base_link;
+
+  /**
+   * Initialize the product module.
+   */
+  public static function init(): void {
+    $obj = new self();
```
bogdanarizancu, line 27: This is not how call methods, please have a look everywhere else, and in my following comment.
```suggestion
```

src/Products.php
```diff
@@ -0,0 +1,136 @@
...
+    if (get_option(self::FIELD_ENFORCE_CAT_LINKS)) {
+      add_filter('post_type_link', [$obj, 'get_product_permalink'], 10, 2);
+    }
+    add_action('admin_init', [$obj, 'register_product_settings']);
+    add_action('admin_init', [$obj, 'save_product_settings']);
```
bogdanarizancu, line 32:
```suggestion
    add_action('admin_init', __CLASS__ . '::register_product_settings');
    add_action('admin_init', __CLASS__ . '::save_product_settings');
```

src/Products.php
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  /**
+   * Registers custom product settings.
+   */
+  public function register_product_settings(): void {
```
bogdanarizancu, line 38: @chrisbaltazar why are adding settings in another place than the Shop Standards tab from WooCommerce settings?

src/Products.php
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  /**
+   * Renders custom option in page.
+   */
+  public function render_permalink_option() { ?>
+      <input name="<?= self::FIELD_ENFORCE_CAT_LINKS ?>" type="checkbox"
+             value="1" <?php
+              checked(get_option(self::FIELD_ENFORCE_CAT_LINKS)) ?> />
+    <?php
```
bogdanarizancu, line 74: What happened with the code formatting here?

src/Products.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    $product_link = str_replace($product_cat_placeholder,
+      $main_term_slug,
+      $base_permalink);
+    $product_link = sprintf('%s/%s',
+      untrailingslashit($product_link),
+      trailingslashit($post->post_name));
+    return home_url($product_link);
+  }
```
bogdanarizancu, line 114: Can you please check your code formatting settings? it's all wrong

src/Products.php
```diff
@@ -0,0 +1,136 @@
...
+    if (isset($this->product_base_link)) {
+      return $this->product_base_link;
+    }
+
+    $product_permalink = get_option('woocommerce_permalinks');
+    if (empty($product_permalink)) {
+      return '';
+    }
+
+    $base_permalink = $product_permalink['product_base'];;
```
bogdanarizancu, line 129: Did you test this somewhere? the double `;;` should crash PHP...
```suggestion
    $base_permalink = $product_permalink['product_base'];
```

bogdanarizancu (changes requested):

src/Plugin.php
chrisbaltazar: I took inspiration from other classes here such as `PlusProducts` which also sounds a bit wide... I'd adopt your suggestion in case you agree to create a special namespace for the whole Product domain classes, otherwise I created this one to start shaping up all possible future product related features from one place

src/Products.php
chrisbaltazar: if we settle down our discussion about the class name itself we can also settle this one ;)

src/Products.php
chrisbaltazar: there are many ways to do this, which one is the right one is the question? abusing the static functions without creating a proper object context is not what I'd do 100% of times... I'd really like to discuss it within the team if is also necessary

bogdanarizancu: @chrisbaltazar — I agree this is debatable in terms of best practices, but we are talking consistency here. If we decide to change the way we do things from now we should all adopt a new way, but until then we should stick to one standard as much as possible.

geist-ahnen (approved):
LGTM

PR #172, opened 2021-11-29
Fixed product filters nofollow links

geist-ahnen:
### Ticket
https://app.asana.com/0/809933051638353/1190953714418427

### Description
Related PR: https://github.com/netzstrategen/gaco/pull/749


src/ProductFilters/WidgetFilterDeliveryTime.php
```diff
@@ -98,7 +98,7 @@ public function widget($args, $instance) {
       }
       $chosen = in_array($delivery_time->term_id, $filter_values, TRUE);
       echo sprintf('<li class="%s">', $chosen ? $item_class . ' ' . $item_chosen_class : $item_class);
-      echo sprintf('<a rel="nofollow" href="%s">%s</a></li>', $link, $delivery_time->name);
+      echo sprintf('<span data-url="%s">%s</span></li>', $link, $delivery_time->name);
```
geist-ahnen, line 101: FYI This change was mentioned in QA by @Henning-Richters as well in the same Asana task

assets/scripts/main.js
```diff
@@ -5,9 +5,9 @@
   $('body')
     .on(
       'click',
-      '.woocommerce-widget-layered-nav-list__item, .widget_layered_nav_filters .chosen, .shop-sidebar-widget .chosen',
+      '[data-url]',
       (e) => {
-        const url = $(e.target).data('url');
+        const url = $(e.target).data('url') || $(e.target).closest('[data-url]').data('url');
```
geist-ahnen, line 10: Sometimes the e.target is a child of `[data-url]`, so if that's the case case we will traverse the DOM to the parent ` $(e.target).closest('[data-url]').data('url');`

chrisbaltazar: not sure this is the right approach, first of all from the docs it says that `closets` "Begins with the current element" so could even just keep this instead of having 2 options here... 
But moreover, this will ONLY look UP and not INTO the children as you say 🤔 ... perhaps what you need is just `$.find()` ?? maybe you can show me one example to better understand this please

chrisbaltazar (changes requested):

assets/scripts/main.js
geist-ahnen: @chrisbaltazar Sure, let me explain this a little better. 
It's actually really simple and a quite common problem in front-end with eventListeners and a common practice.

The e.target is the actual element that gets clicked, 

So if you have an HTML structure as:
```
<div data-url="https://google.com"> 
 <img src="my-cool-image"/>  
</div>
```
And you click in the `img` and you try to get from` e.target.data('url')` you will get undefined, because the img tag doesn't contain any data-url attribute and it is the e.target. So to prevent that you need to traverse the DOM up, never down.

In case that e.target was in fact the element with the `data-url attribute` then this part of the code will not run `|| $(e.target).closest('[data-url]').data('url')`

It is called short circuit evaluation: https://codeburst.io/javascript-what-is-short-circuit-evaluation-ff22b2f5608c

This why not matter if we click the parent or the child we will get the data-url.

chrisbaltazar: all clear now thanks

chrisbaltazar (approved):
💯

PR #174, opened 2021-12-03
Removed email & billing address checkout validation in AmazonV2Pay.

geist-ahnen:
### Ticket
https://app.asana.com/0/1201154968324561/1201075091917011/f

### DIFF
--- a/src/WooCommerceCheckout.php
+++ b/src/WooCommerceCheckout.php
@@ -15,7 +15,7 @@ class WooCommerceCheckout {
     // Removes city field from shipping calculator.
     add_filter('woocommerce_shipping_calculator_enable_city', '__return_false');
 
-    if (is_admin() || wp_doing_cron()) {
+    if (is_admin()) {
       return;
     }
 
@@ -34,7 +34,9 @@ class WooCommerceCheckout {
     add_filter( 'woocommerce_form_field', __CLASS__ . '::woocommerceFormField', 10, 4 );
 
     // Remove required fields while using Amazon Pay  V2.
-    add_filter('woocommerce_checkout_fields', __CLASS__ . '::removeRequiredFieldsforAmazonPay', 5);
+    if (self::isAmazonPayV2Checkout()) {
+      add_filter('woocommerce_checkout_fields', __CLASS__ . '::removeRequiredFieldsforAmazonPay');
+    }
   }
 
   /**
@@ -130,7 +132,7 @@ class WooCommerceCheckout {
   public static function isAmazonPayV2Checkout(): bool {
     $is_amzon_pay_active = is_plugin_active('woocommerce-gateway-amazon-payments-advanced/woocommerce-gateway-amazon-payments-advanced.php') ?: FALSE;
     if ($is_amzon_pay_active && isset(WC()->session)) {
-      if(version_compare(WC_AMAZON_PAY_VERSION, '2.0', '>=') && WC()->session->get('amazon_checkout_session_id') !== null) {
+      if (defined('WC_AMAZON_PAY_VERSION') && version_compare(WC_AMAZON_PAY_VERSION, '2.0', '>=') && !empty(WC()->session->get('amazon_checkout_session_id'))) {
         return true;
       }
     }
@@ -138,18 +140,14 @@ class WooCommerceCheckout {
   }
 
   /**
-   * Remove Billing address when using Amazon Pay V2.
-   *
-   * * @return array
-   *   checkout fields.
+   * Remove required fields when checking out via Amazon using
+   * woocommerce-gateway-amazon-payments-advanced.
    * 
    * @implements woocommerce_checkout_fields
    */
   public static function removeRequiredFieldsforAmazonPay(array $fields): array {
-    if (self::isAmazonPayV2Checkout()) {
-        unset($fields['billing']['billing_address_1']);
-        return $fields;
-    }
+    unset($fields['billing']['billing_address_1']);
     return $fields;
   }
+
 }
--- a/src/WooCommerceCheckout.php
+++ b/src/WooCommerceCheckout.php
@@ -23,7 +23,8 @@ class WooCommerceCheckout {
     if (
       get_option('_' . Plugin::L10N . '_checkout_email_confirmation_field') === 'yes' &&
       !is_user_logged_in() &&
-      !isset($_GET['woo-paypal-return'])
+      !isset($_GET['woo-paypal-return']) &&
+      !self::isAmazonPayV2Checkout()
     ) {
       add_filter('woocommerce_checkout_fields', __CLASS__ . '::addConfirmationEmailCheckoutField');
       add_action('woocommerce_checkout_process', __CLASS__ . '::checkConfirmationEmailField');
@@ -31,6 +32,9 @@ class WooCommerceCheckout {
 
     // Add checkout error messages.
     add_filter( 'woocommerce_form_field', __CLASS__ . '::woocommerceFormField', 10, 4 );
+
+    // Remove required fields while using Amazon Pay  V2.
+    add_filter('woocommerce_checkout_fields', __CLASS__ . '::removeRequiredFieldsforAmazonPay', 5);
   }
 
   /**
@@ -118,4 +122,34 @@ class WooCommerceCheckout {
     }
   }
 
+  /**
+   * Check if we are using Amazon V2 in the checkout.
+   *
+   * @return bool
+   */
+  public static function isAmazonPayV2Checkout(): bool {
+    $is_amzon_pay_active = is_plugin_active('woocommerce-gateway-amazon-payments-advanced/woocommerce-gateway-amazon-payments-advanced.php') ?: FALSE;
+    if ($is_amzon_pay_active) {
+      if(version_compare(WC_AMAZON_PAY_VERSION, '2.0', '>=') && WC()->session->get('amazon_checkout_session_id') !== null) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Remove Billing address when using Amazon Pay V2.
+   *
+   * * @return array
+   *   checkout fields.
+   * 
+   * @implements woocommerce_checkout_fields
+   */
+  public static function removeRequiredFieldsforAmazonPay(array $fields): array {
+    if (self::isAmazonPayV2Checkout()) {
+        unset($fields['billing']['billing_address_1']);
+        return $fields;
+    }
+    return $fields;
+  }
 }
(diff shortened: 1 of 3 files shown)


src/WooCommerceCheckout.php
```diff
@@ -118,4 +122,34 @@ public static function checkConfirmationEmailField() {
...
 
+  /**
+   * Check if we are using Amazon V2 in the checkout.
+   *
+   * @return bool
+   */
+  public static function isAmazonPayV2Checkout(): bool {
+    $is_amzon_pay_active = is_plugin_active('woocommerce-gateway-amazon-payments-advanced/woocommerce-gateway-amazon-payments-advanced.php') ?: FALSE;
+    if ($is_amzon_pay_active) {
+      if(version_compare(WC_AMAZON_PAY_VERSION, '2.0', '>=') && WC()->session->get('amazon_checkout_session_id') !== null) {
```
geist-ahnen, line 133: I am using the constant `WC_AMAZON_PAY_VERSION` provided by the plugin itself instead of getting it with `get_plugin_data()` because their plugin header doesn't adjust to WordPress standards and it was not returning the correct info: https://developer.wordpress.org/reference/functions/get_plugin_data/

bogdanarizancu (approved):

src/WooCommerceCheckout.php
sun, line 133: This causes a fatal error on every cron run; see https://app.tideways.io/o/netzstrategen/gaco-stage/issues/errors?env=production&s=web&error=11305-902b0d55fddef6f8d651fe1035b7d4bd

image

sun (changes requested):

src/WooCommerceCheckout.php
geist-ahnen: `WC()->session` is being shown as undefined, probably because there is not a session in cron.

I should verify first it with  `isset(WC()->session)` or even `!WC()->session->has_session()`

What you think @sun ?

src/WooCommerceCheckout.php
```diff
@@ -15,22 +15,26 @@ public static function init() {
     // Removes city field from shipping calculator.
     add_filter('woocommerce_shipping_calculator_enable_city', '__return_false');
 
-    if (is_admin()) {
+    if (is_admin() || wp_doing_cron()) {
```
geist-ahnen, line 18: This file is only related to the WooCommerce Checkout so we can just return if we are doing a cron job.

src/WooCommerceCheckout.php
```diff
@@ -118,4 +122,34 @@ public static function checkConfirmationEmailField() {
...
   }
 
+  /**
+   * Check if we are using Amazon V2 in the checkout.
+   *
+   * @return bool
+   */
+  public static function isAmazonPayV2Checkout(): bool {
+    $is_amzon_pay_active = is_plugin_active('woocommerce-gateway-amazon-payments-advanced/woocommerce-gateway-amazon-payments-advanced.php') ?: FALSE;
+    if ($is_amzon_pay_active && isset(WC()->session)) {
```
geist-ahnen, line 132: To avoid errors, I am also verifying if the WC()->session is set.

src/WooCommerceCheckout.php
sun: I understand the thinking but I would not advise to do this, because if we are going to add some checkout related customization in the future, then we will debug a long time why the system behaves differently during cron runs.
```suggestion
    if (is_admin()) {
```

src/WooCommerceCheckout.php
```diff
@@ -15,22 +15,26 @@ public static function init() {
...
     ) {
       add_filter('woocommerce_checkout_fields', __CLASS__ . '::addConfirmationEmailCheckoutField');
       add_action('woocommerce_checkout_process', __CLASS__ . '::checkConfirmationEmailField');
     }
 
     // Add checkout error messages.
     add_filter( 'woocommerce_form_field', __CLASS__ . '::woocommerceFormField', 10, 4 );
+
+    // Remove required fields while using Amazon Pay  V2.
+    add_filter('woocommerce_checkout_fields', __CLASS__ . '::removeRequiredFieldsforAmazonPay', 5);
```
sun, line 37: 1. We should not register this filter callback in the first place if the plugin is not active (leaving only the session condition within the callback function).
2. Can you add a comment (above the line) to explain the special priority 5?

src/WooCommerceCheckout.php
```diff
@@ -15,22 +15,26 @@ public static function init() {
...
+      !self::isAmazonPayV2Checkout()
     ) {
       add_filter('woocommerce_checkout_fields', __CLASS__ . '::addConfirmationEmailCheckoutField');
       add_action('woocommerce_checkout_process', __CLASS__ . '::checkConfirmationEmailField');
     }
 
     // Add checkout error messages.
     add_filter( 'woocommerce_form_field', __CLASS__ . '::woocommerceFormField', 10, 4 );
+
+    // Remove required fields while using Amazon Pay  V2.
```
sun, line 36: We understand the context right now, because we're working on it, but it's going to be harder in one year from now.
```suggestion
    // Remove required fields when checking out via Amazon using
    // woocommerce-gateway-amazon-payments-advanced.
```

src/WooCommerceCheckout.php
sun: This is the actual required fix for my past comment https://github.com/netzstrategen/wordpress-shop-standards/pull/174/files#r765024107

src/WooCommerceCheckout.php
sun: That's okay, but you should check whether the constant is defined before using it in order to prevent a PHP warning in the future in case the plugin changes its constants.

It seems like you're not actually interested in the content, only whether the session parameter is set.  Therefore, you should use `isset()` or `!empty()` for this instead of comparing against `null`.

Doing this should also allow you to simplify and shorten the code to a single `isset()`, because PHP evaluates the expression from left to right and returns a negative result for the whole `isset()` expression if any property or value in the chain is not set.

Please also note that your code is not following our coding standards in various places (here: missing space after `if`).
```suggestion
      if (defined('WC_AMAZON_PAY_VERSION') && version_compare(WC_AMAZON_PAY_VERSION, '2.0', '>=') && isset(WC()->session->get('amazon_checkout_session_id'))) {
```

src/WooCommerceCheckout.php
```diff
@@ -118,4 +122,34 @@ public static function checkConfirmationEmailField() {
...
+    }
+    return false;
+  }
+
+  /**
+   * Remove Billing address when using Amazon Pay V2.
+   *
+   * * @return array
+   *   checkout fields.
+   *
```
sun, line 145: Minor: We don't document the parameters and return value of a function if it is a defined hook that is already documented elsewhere (as indicated by `@implements`).
```suggestion
```
(Note that there was a duplicate asterisk before `@return`.)

src/WooCommerceCheckout.php
```diff
@@ -118,4 +122,34 @@ public static function checkConfirmationEmailField() {
...
+   * * @return array
+   *   checkout fields.
+   * 
+   * @implements woocommerce_checkout_fields
+   */
+  public static function removeRequiredFieldsforAmazonPay(array $fields): array {
+    if (self::isAmazonPayV2Checkout()) {
+        unset($fields['billing']['billing_address_1']);
+        return $fields;
```
sun, line 151: Minor: Wrong indentation.
```suggestion
      unset($fields['billing']['billing_address_1']);
      return $fields;
```
(Our linter automations should eliminate the need for such manual comments in the future. 🙏 )

src/WooCommerceCheckout.php
```diff
@@ -118,4 +122,34 @@ public static function checkConfirmationEmailField() {
...
+    }
+    return $fields;
+  }
```
sun, line 154: Minor: There should be a single blank line between function declarations and before the first and after the last.
```suggestion
  }

```

sun (commented):
@geist-ahnen 👍  Looks better now – but after having an actual look through the code of this PR now (I only looked at the line causing the fatal error before), here's some more feedback for you 🙂

src/WooCommerceCheckout.php
geist-ahnen: Makes sense. Thanks for pointing it out.

src/WooCommerceCheckout.php
geist-ahnen: Noted. :)

src/WooCommerceCheckout.php
geist-ahnen: @sun "It seems like you're not actually interested in the content, only whether the session parameter is set. Therefore, you should use isset() or !empty() for this instead of comparing against null."

You are correct, I am not interested in the content, but without doing that comparison I get this error, so I have left it as it. Any advise?

`It seems like you're not actually interested in the content, only whether the session parameter is set. Therefore, you should use isset() or !empty() for this instead of comparing against null.`

src/WooCommerceCheckout.php
```diff
@@ -118,4 +124,30 @@ public static function checkConfirmationEmailField() {
...
+      if (defined('WC_AMAZON_PAY_VERSION') && version_compare(WC_AMAZON_PAY_VERSION, '2.0', '>=') && !empty(WC()->session->get('amazon_checkout_session_id'))) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Remove required fields when checking out via Amazon using
+   * woocommerce-gateway-amazon-payments-advanced.
```
sun, line 144:
```suggestion
   * Remove required fields when checking out via Amazon using woocommerce-gateway-amazon-payments-advanced.
```

src/WooCommerceCheckout.php
```diff
@@ -118,4 +125,29 @@ public static function checkConfirmationEmailField() {
...
   }
 
+  /**
+   * Check if we are using Amazon V2 in the checkout.
+   *
+   * @return bool
+   */
+  public static function isAmazonPayV2Checkout(): bool {
+    $is_amzon_pay_active = is_plugin_active('woocommerce-gateway-amazon-payments-advanced/woocommerce-gateway-amazon-payments-advanced.php') ?: FALSE;
+    if ($is_amzon_pay_active && isset(WC()->session)) {
```
sun, line 135: 1. is_plugin_active() only returns true or false, so the ternary is unnecessary.
2. Typo in "amzon".
```suggestion
    $is_amazon_pay_active = is_plugin_active('woocommerce-gateway-amazon-payments-advanced/woocommerce-gateway-amazon-payments-advanced.php');
    if ($is_amazon_pay_active && isset(WC()->session)) {
```

sun (approved):

PR #175, opened 2021-12-20
Fixed condition to change only published product links allowing draft previews.

chrisbaltazar:
### Ticket
https://app.asana.com/0/1200981693019240/1201538907448298

### Description
We missed a detail on the previous fix which is now affecting draft previews, so now it will only work for published products as intented.


geist-ahnen (approved):
LGTM

PR #177, opened 2021-12-27
Added checkbox for defective or used goods.

Mauricio-Urrego:
### Ticket
- [Checkbox for defective or used goods](https://app.asana.com/0/30156186242982/1201508886861217/f)

### Description
The new request for this task is to rely ONLY on a single product attribute (at the moment called "condition") in order to know when is needed to display the new checkbox before the cart button. But the plugin is ready to configure and change this property and it's value when needed as well.


src/ProductDefects.php
```diff
@@ -0,0 +1,239 @@
+<?php
+
+/**
+ * @file
+ * Contains \Netzstrategen\ShopStandards\UsedGoods.
```
bogdanarizancu, line 5: I think we need to decide if it is `UsedGoods` or `ProductDefects`, to me the first option sounds more descriptive, in any case it should be one name in the comments, class and file names.

src/ProductDefects.php
```diff
@@ -0,0 +1,239 @@
...
+  /**
+   * Checkbox consent initialization method.
+   */
+  public static function init() {
+    // Logs consent to defective or used product when product is added to cart.
+    add_action('woocommerce_add_to_cart', function () {
+      ProductDefects::logConsent();
+    });
+
+    add_action('wp', function () {
```
bogdanarizancu, line 24: why is this an anonymous function and not a static method, like the rest?

src/ProductDefects.php
```diff
@@ -0,0 +1,239 @@
...
+      $show_value = get_post_meta(get_the_ID(), '_' . Plugin::PREFIX . '_show_product_defects_consent');
+      if ($show_value && $show_value[0] === 'yes') {
+        $show_override = TRUE;
+      }
+      else {
+        $show_override = FALSE;
+      }
```
bogdanarizancu, line 31: In my opinion this can be written shorter, like:
```suggestion
     $show_override = $show_value && $show_value[0] === 'yes' ? TRUE : FALSE;
```

bogdanarizancu (commented):
@Mauricio-Urrego — nice work, just a few preliminary comments since this is still in Draft mode

src/ProductDefects.php
sun: Although UsedGoods might be more generic in natural language, I would recommend to go with ProductDefects, because it contains more relevant keywords in the code space.

Mauricio-Urrego: In any case this was already defined in the ticket description. I forgot to change this one (it started with naming everything UsedGoods until I noticed).

src/ProductDefects.php
Mauricio-Urrego: Type. Thanks!

src/ProductDefects.php
Mauricio-Urrego: I like the ternaries but I thought it might be clearer this way. I'll change it because well, I don't have strong opinions.

chrisbaltazar: How the hell will not work 😄 ... I just came by to add my own final suggestion which is just short this as 
```
$show_override = $show_value && $show_value[0] === 'yes';
```
and that should do it 😉

src/Plugin.php
```diff
@@ -113,6 +113,7 @@ public static function init() {
     WooCommerceCheckout::init();
     PlusProducts::init();
     ProductsPermalinks::init();
+    ProductDefects::init();
```
bogdanarizancu, line 116: Should we wrap this inside a check so that it would only fire on the single product template/URL? I think that would save it from running `ProductDefects::display_product_checkbox` on each request.

src/ProductDefects.php
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    $uploads_dir = wp_upload_dir()['basedir'] . '/' . Plugin::PREFIX;
+    $log_file = $uploads_dir . '/defects-consent.log';
+    if (!dir($uploads_dir)) {
+      wp_mkdir_p($uploads_dir);
+    }
+    $data = [
+      'timestamp' => date_i18n('c'),
+      'user' => get_current_user_id(),
```
bogdanarizancu, line 85: Do we log this if the customer is not logged in?

src/ProductDefects.php
```diff
@@ -0,0 +1,173 @@
...
+   *
+   * @implemements woocommerce_available_variation
+   */
+  public static function pass_variation_attribute($data, $product, $variation) {
+    $data['used_goods_consent_attribute'] = self::get_display_product_attribute($variation);
+    return $data;
+  }
+
+  /**
+   * Adds checkout specific settings.
```
bogdanarizancu, line 127: Is this doc right?

bogdanarizancu (approved):
@chrisbaltazar this is good work, just a few minor comments/questions, otherwise approving this already.

src/Plugin.php
chrisbaltazar: Good remark, in this case and the majority of them I'd handle that logic into each module initializer so we keep this as a bootstrapper point, moreover the timing to add that condition is too early here and it won't work (I tried it) but thanks to your idea I improved the condition to bail from the method as soon as is called if is not supposed to run

src/ProductDefects.php
chrisbaltazar: Frankly I just left this part pretty much untouched but I don't know the reason behind this, nothing written in the ticket and it was not mentioned by the PM's when discussed this, so I will better get rid of it

src/ProductDefects.php
chrisbaltazar: good catch

chrisbaltazar (commented):
Thanks @bogdanarizancu I will address your comments now

PR #182, opened 2022-01-26
Added filter hook to add shipping method to imported Amazon orders.

geist-ahnen:
### Ticket
https://app.asana.com/0/30156186242982/1192093200609628/f

### DIFF
--- a/src/Amazon.php
+++ b/src/Amazon.php
@@ -8,11 +8,11 @@ namespace Netzstrategen\ShopStandards;
 class Amazon {
 
   /**
-   * First Available Shipping Method or null.
+   * The first available shipping method keyed by order ID.
    *
-   * @var WC_Shipping_Rate|null
+   * @var WC_Shipping_Rate[]
    */
-  public static $firstAvailableMethod = NULL;
+  public static $firstAvailableMethod = [];
 
   /**
    * Replaces the order ID of imported Amazon orders with the custom order ID.
@@ -32,7 +32,7 @@ class Amazon {
    * Returns whether we are using Amazon Pay v2.
    *
    * @return bool
-   *   True or false depending is Amazon Pay v2 is active.
+   *   True or false depending on whether Amazon Pay v2 is active.
    */
   public static function isAmazonPayV2Checkout(): bool {
     $is_amazon_pay_active = is_plugin_active('woocommerce-gateway-amazon-payments-advanced/woocommerce-gateway-amazon-payments-advanced.php');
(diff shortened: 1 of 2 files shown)


src/Amazon.php
```diff
@@ -7,9 +7,16 @@
  */
 class Amazon {
 
+  /**
+   * First Available Shipping Method or null.
+   *
+   * @var WC_Shipping_Rate|null
+   */
+  public static $firstAvailableMethod = NULL;
```
sun, line 15: If multiple orders are imported in one process, then the same shipping method would be applied to all of them. This should be keyed by order ID.
```suggestion
   * @var WC_Shipping_Rate[]
   */
  public static $firstAvailableMethod = [];
```

src/Amazon.php
```diff
@@ -25,13 +32,105 @@ public static function woocommerce_amazon_pa_update_checkout_session_payload($pa
...
+    // Calculate shipping.
+    $packages = $cart->get_shipping_packages();
+    $shipping->calculate_shipping($packages);
+    $available_methods = $shipping->get_packages();
+
+    if (!empty($available_methods)) {
+      self::$firstAvailableMethod = current($available_methods[0]['rates']);
+    }
+
+    return self::$firstAvailableMethod;
```
sun, line 133:
```suggestion
    return self::$firstAvailableMethod[$order->get_id()];
```

sun (approved):
Still very good. I found some issues on a second look; especially the missing key by order ID needs to be addressed. Feel free to apply all suggestions directly from the PR, but please ensure to test again. Approving already, so you can merge yourself when done. 👍

src/Amazon.php
```diff
@@ -25,13 +32,105 @@ public static function woocommerce_amazon_pa_update_checkout_session_payload($pa
...
+    // Calculate shipping.
+    $packages = $cart->get_shipping_packages();
+    $shipping->calculate_shipping($packages);
+    $available_methods = $shipping->get_packages();
+
+    if ($available_methods) {
+      self::$firstAvailableMethod[$order->get_id()] = reset($available_methods[0]['rates']);
+    }
+
+    return self::$firstAvailableMethod[$order->get_id()];
```
sun, line 133: Triggers a PHP notice/warning if there are no available methods.

```diff
    if ($available_methods) {
      self::$firstAvailableMethod[$order->get_id()] = reset($available_methods[0]['rates']);
    }
+   else {
+     return FALSE;
+   }

    return self::$firstAvailableMethod[$order->get_id()];
```

Or if the calculated result should be sticky even if there are no available methods:
```diff
    if ($available_methods) {
      self::$firstAvailableMethod[$order->get_id()] = reset($available_methods[0]['rates']);
    }
+   else {
+     self::$firstAvailableMethod[$order->get_id()] = FALSE;
+   }

    return self::$firstAvailableMethod[$order->get_id()];
```

PR #186, opened 2022-02-10
Fixed fatal error due to missing session when creating cart programmatically outside frontend.

geist-ahnen:


src/Amazon.php
```diff
@@ -134,6 +149,10 @@ public static function get_available_shipping_methods(?\WC_Order $order) {
       self::$firstAvailableMethod[$order->get_id()] = reset($available_methods[0]['rates']);
     }
 
+    // Clean up
```
geist-ahnen, line 152: I am considering also emptying the cart as part of our clean up.

src/Amazon.php
```diff
@@ -113,8 +115,21 @@ public static function get_available_shipping_methods(?\WC_Order $order) {
       return FALSE;
     }
 
+    if (!did_action('woocommerce_load_cart_from_session')) {
+      $GLOBALS['wp_actions']['woocommerce_load_cart_from_session'] = TRUE;
+    }
```
sun, line 120: As you're creating an actual session now, this hack shouldn't be necessary anymore?

src/Amazon.php
sun: Yeah, emptying the cart would be good to avoid it from starting another session on shutdown.

src/Amazon.php
geist-ahnen: I will test it without this and remove it if needed.

geist-ahnen: Jup, it worked well without this.

PR #188, opened 2022-03-03
Fixed cart validation failure always disables button for checkout.

bogdanarizancu:
### Ticket
- [Shop Standards Plugin: on the Cart Page wrong coupon code prevents that customers can check out ](https://app.asana.com/0/809933051638353/1201450852483666)

### Related commit
- https://github.com/netzstrategen/wordpress-shop-standards/commit/f5db983f551385814d5c420c72063718db588d9c


assets/scripts/main.js
```diff
@@ -102,7 +102,7 @@
 
   // Disable checkout button if there are any WooCommerce error displayed.
   $(document).on('ready updated_cart_totals', () => {
-    if ($('.woocommerce-error').length) {
+    if ($('.woocommerce-error li[data-plus-product="invalid"]').length) {
```
sun, line 105: Did you intend to mean the opposite here; i.e., if the error is NOT the plus-product=invalid error?

The proposed change essentially breaks the desired functionality of removing the checkout button in case of error messages in the checkout, which can't be right?  If that was our intention, we would rather remove this whole block altogether…?

sun (changes requested):

assets/scripts/main.js
bogdanarizancu: @sun — no, I did not mean that at all. If you take a look at the referenced initial commit of this PR's description, you can understand why this was initially wrongly implemented, as the checkout button should have been disabled only if a plus product was added to the cart and not in any other condition. However the JS here checks for any `woocommerce-error` regardless and disables checkout if any is found. This is an extreme and completely not recommended solution (you would almost never would want to prevent a customer from completing an order, in any case, not for entering an inexistent coupon code). So my thinking here was the single plus product in cart is the exception, and only then would we want to prevent completing an order.

sun: Okay, makes sense then.  Can you update the comment on line 103 accordingly?

sun (approved):

PR #191, opened 2022-03-14
Added product inventory incoming stock field.

bogdanarizancu:
### Ticket
- [Additional backend field "incoming stock"](https://app.asana.com/0/30156186242982/1201791757841954/f)


languages/shop-standards-de_DE.po
```diff
@@ -171,110 +179,114 @@ msgstr ""
 msgid "Shop Standards"
 msgstr ""
 
-#: src/WooCommerce.php:29
+#: src/WooCommerce.php:35 src/WooCommerce.php:680
+msgid "Marketing focus product"
+msgstr "Marketing focus product"
```
fabianmarz, line 184: minor: If a translation is the same it can be omitted

src/Admin.php
```diff
@@ -22,6 +22,10 @@ class Admin {
   public static function init() {
     // Adds custom fields for single products.
     add_action('woocommerce_product_options_general_product_data', __NAMESPACE__ . '\WooCommerce::woocommerce_product_options_general_product_data');
+    // Adds incoming stock WooCommerce to simple products inventory.
+    add_action('woocommerce_product_options_stock_fields', __NAMESPACE__ . '\WooCommerce::showIncomingstock', 9);
+    // Adds incoming stock WooCommerce to variable products inventory.
+    add_action('woocommerce_product_after_variable_attributes', __NAMESPACE__ . '\WooCommerce::showIncomingstock', 9, 3);
```
fabianmarz, line 28: I don't understand that comment. 🤔 

Also since two hooks are doing the same I'd rather write a single comment
```suggestion
    // Adds product option to show incoming stock.
    add_action('woocommerce_product_options_stock_fields', __NAMESPACE__ . '\WooCommerce::showIncomingstock', 9);
    add_action('woocommerce_product_after_variable_attributes', __NAMESPACE__ . '\WooCommerce::showIncomingstock', 9, 3);
    
```
Is the 2 argument needed in the latter hook? The method seems not to support any parameters.

src/WooCommerce.php
```diff
@@ -388,6 +389,27 @@ public static function woocommerce_product_options_general_product_data() {
...
+   */
+  public static function showIncomingstock() {
+    if (ProductFieldsManager::show_field(self::FIELD_PRODUCT_INCOMING_STOCK)) {
+      // Back in stock date field.
+      echo '<div class="options_group show_if_simple show_if_external">';
+      woocommerce_wp_text_input([
+        'id'          => self::FIELD_PRODUCT_INCOMING_STOCK,
+        'label'       => __('Incoming Stock', Plugin::L10N),
+        'desc_tip'    => 'true',
+        'description' => self::get_product_fields()[self::FIELD_PRODUCT_INCOMING_STOCK],
```
fabianmarz, line 407: Seems the checkbox in `woocommerce_product_options_related` is using wrong formatting as well, but overall I think this file uses our usual conventions.

```suggestion
        'id' => self::FIELD_PRODUCT_INCOMING_STOCK,
        'label' => __('Incoming Stock', Plugin::L10N),
        'desc_tip'  => 'true',
        'description' => self::get_product_fields()[self::FIELD_PRODUCT_INCOMING_STOCK],
```

src/Admin.php
bogdanarizancu: @fabianmarz I had to move the second one, as requested by Stephan to show in different order than what was possible like so, please review again.

PR #192, opened 2022-03-17
Fixed street is missing in billing address for German orders paid with Amazon Pay.

geist-ahnen:
### Ticket
https://app.asana.com/0/30156186242982/1201940215532388/f

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.42.3",
+  "version": "1.42.4",
   "title": "shop-standards",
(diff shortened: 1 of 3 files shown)


sun (commented):
Didn't you have additional changes to transfer the value from line 2 to 1 if only in there?

geist-ahnen:
> Didn't you have additional changes to transfer the value from line 2 to 1 if only in there?

That fix is not longer needed. Stephan provided more information in the QA ticket, and it is actually a bug on their side.
I will create a patch for this and also send it upstream.

PR #198, opened 2022-04-07
Allow to add multiple products to the cart.

juanlopez4691:
### Ticket
- [Allow to add more than one product to cart via URL](https://app.asana.com/0/1199938965307042/1202031768897952/f)

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -741,6 +741,56 @@ class WooCommerce {
   }
 
   /**
+   * Allows adding multiple products to the cart.
+   *
+   * See https://bit.ly/3JfcIET
+   *
+   * @implements wp_loaded
+   */
+  public static function add_multiple_products_to_cart() {
+    if (
+      !isset($_REQUEST['add-to-cart']) ||
+      strpos($_REQUEST['add-to-cart'], ',') === false
+    ) {
+      return;
+    }
+
+    remove_action('wp_loaded', array('WC_Form_Handler', 'add_to_cart_action'), 20);
+
+    $product_ids = explode(',', $_REQUEST['add-to-cart']);
+
+    // Avoid redirection to cart after adding a product.
+    $fn_return_no = fn() => 'no';
+    add_filter('woocommerce_add_to_cart_redirect', '__return_false', 99);
+    add_filter('option_woocommerce_cart_redirect_after_add', $fn_return_no, 99);
+
+    array_map(
+      function ($product_id) {
+        $product_id = apply_filters('woocommerce_add_to_cart_product_id', absint($product_id));
+        if (!$product = wc_get_product($product_id)) {
+          return;
+        }
+
+        $handler = apply_filters('woocommerce_add_to_cart_handler', $product->get_type(), $product);
+
+        if ($handler === 'variable' || $handler === 'variation') {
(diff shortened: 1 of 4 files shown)


src/WooCommerce.php
```diff
@@ -740,6 +740,52 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
...
+
+    remove_action('wp_loaded', array('WC_Form_Handler', 'add_to_cart_action'), 20);
+
+    $product_ids = explode(',', $_REQUEST['add-to-cart']);
+    if (!$product_ids) {
+      return;
+    }
+    $num_products = count($product_ids);
+
+    $adding_to_cart = TRUE;
```
bogdanarizancu, line 766: This doesn't seem to be used anywhere.
```suggestion
```

src/WooCommerce.php
```diff
@@ -740,6 +740,52 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
...
+    ) {
+      return;
+    }
+
+    remove_action('wp_loaded', array('WC_Form_Handler', 'add_to_cart_action'), 20);
+
+    $product_ids = explode(',', $_REQUEST['add-to-cart']);
+    if (!$product_ids) {
+      return;
+    }
```
bogdanarizancu, line 763: @juanlopez4691 — I think explode would always return at least 1 id, this check is not needed here in my opinion
```suggestion
```

src/WooCommerce.php
```diff
@@ -740,6 +740,52 @@ public static function woocommerce_save_product_variation($post_id, $loop) {
...
+    $redirect_to_cart = get_option('woocommerce_cart_redirect_after_add');
+
+    // Avoid redirection to cart after adding a product.
+    $fn_return_no = fn() => 'no';
+    add_filter('woocommerce_add_to_cart_redirect', '__return_false', 99);
+    add_filter('option_woocommerce_cart_redirect_after_add', $fn_return_no, 99);
+
+    foreach ($product_ids as $index => $product_id) {
+      $product_id = apply_filters('woocommerce_add_to_cart_product_id', absint($product_id));
+      if (empty($product_id)) {
```
bogdanarizancu, line 776: I just found a similar implementation for AJAX calls at https://github.com/netzstrategen/wordpress-shop-styleboards/blob/c82b63b2a56dc57a2f4622a8e41caf38dcad2f7e/src/Plugin.php#L644-L679, I think we should validate first before adding to cart (have not tested this code).
```suggestion
      $passed_validation = apply_filters('woocommerce_add_to_cart_validation', TRUE, $product_id, 1);
      if ($passed_validation) {
```

bogdanarizancu (changes requested):

src/WooCommerce.php
juanlopez4691: I prefer to keep it, just in case no product ID at all is added to the URL. Just to be defensive.

src/WooCommerce.php
juanlopez4691: AFAIK hook `woocommerce_add_to_cart_validation` is used to allow custom validation when adding a product to the cart. I think is a good idea to use it, yes.

bogdanarizancu (approved):
👍🏼

PR #206, opened 2022-05-31
Fixed variant information not changing when other variant is selected.

alinistrate:
### Ticket
- [Bug: Swatches : pre-defined variant information doesn't change, when other variant is selected](https://app.asana.com/0/1199938965307042/1202304267734587)


bogdanarizancu (approved):

bogdanarizancu (approved):

assets/scripts/main.js
```diff
@@ -42,6 +42,12 @@
     $variationSelectChanged = $(this);
   });
 
+  $('.lowest_delivery_variations__select').on('change', function() {
+    setTimeout(function(){
+      $('.variations_form select').trigger('change');
+    }, 100);
+  });
+
```
bogdanarizancu, line 50:
```suggestion
```
We usually tend to avoid `setTimeout` as it actually pauses the whole JS execution for the specified duration, any idea why we would need wrapped in the `timeout`? Is it working without it?

bogdanarizancu (changes requested):

assets/scripts/main.js
alinistrate: We haven't find any other way to make this work without using the timeout. 
We use it because the value of the `$('.variations_form select')` element is not changed until the 100ms timeout. I think that the problem comes from the timeout set for the `check_variations` event on line 69.

bogdanarizancu: Have you considered trying `addEventListener` for the `check_variations` event instead? Something like:
```
addEventListener('check_variations', function() {
  $('.variations_form select').trigger('change'));
});
```

PR #213, opened 2022-08-05
Added delete orphans variants cron event.

creative-andrew:
### Ticket
https://app.asana.com/0/0/1202584852391692/f

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -41,7 +41,7 @@ class Plugin {
   const CRON_EVENT_REMOVE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
 
   /**
- * Cron event name for delete orphan variants.
+   * Cron event name for delete orphan variants.
    *
    * @var string
    */
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -41,6 +41,13 @@ class Plugin {
   const CRON_EVENT_REMOVE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
 
   /**
+ * Cron event name for delete orphan variants.
+   *
+   * @var string
+   */
+  const CRON_EVENT_ORPHANS_VARIANTS_CLEANUP = Plugin::PREFIX . '/cron/orphans-variant-cleanup';
+
+  /**
    * Plugin initialization method with the lowest possible priority.
    *
    * @implements init
@@ -210,6 +217,13 @@ class Plugin {
         wp_schedule_event(strtotime('03:00:00'), 'daily', self::CRON_EVENT_REMOVE_BACK_IN_STOCK);
       }
     }
+
+    // Schedules delete orphan variants cron.
+    if (!wp_next_scheduled(self::CRON_EVENT_ORPHANS_VARIANTS_CLEANUP)) {
+      wp_schedule_event(time(), 'weekly', self::CRON_EVENT_ORPHANS_VARIANTS_CLEANUP);
+    }
+
+    add_action(self::CRON_EVENT_ORPHANS_VARIANTS_CLEANUP, __NAMESPACE__  . '\WooCommerce::cron_orphans_variants_cleanup');
 
     // Prefetches DNS entries for particular resources.
     add_filter('wp_resource_hints', __NAMESPACE__ . '\Performance::wp_resource_hints', 10, 2);
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -45,7 +45,7 @@ class Plugin {
    *
    * @var string
    */
-  const CRON_EVENT_ORPHANS_VARIANTS_CLEANUP = Plugin::PREFIX . '/cron/orphans-variant-cleanup';
+  const CRON_EVENT_ORPHAN_VARIANTS_CLEANUP = Plugin::PREFIX . '/cron/orphan-variants-cleanup';
 
   /**
    * Plugin initialization method with the lowest possible priority.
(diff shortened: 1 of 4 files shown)


src/WooCommerce.php
```diff
@@ -249,6 +249,19 @@ public static function cron_remove_back_in_stock() {
     }
   }
 
+  /**
+   * Cron event callback to remove outdated back-in-stock product metadata.
+   */
+  public static function cron_orphans_variants_cleanup() {
+    global $wpdb;
+    $wpdb->query(
```
creative-andrew, line 257: Query taken of out the method use by the wc cli: https://github.com/woocommerce/woocommerce/blob/3a76aa42555f037a0a565e9f0feaa43f89b197d2/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-system-status-tools-v2-controller.php#L464

We could also instantiate` WC_REST_System_Status_Tools_V2_Controller` can call execute_tool(''delete_orphaned_variations")

src/Plugin.php
```diff
@@ -211,6 +218,13 @@ public static function init() {
       }
     }
 
+    // Schedules delete orphan variants cron.
```
creative-andrew, line 221: Apparently we are putting our crons here. 
Click the arrow to see other crons added in the past.

src/Plugin.php
```diff
@@ -40,6 +40,13 @@ class Plugin {
    */
   const CRON_EVENT_REMOVE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
 
+  /**
+ * Cron event name for delete orphan variants.
+   *
+   * @var string
+   */
+  const CRON_EVENT_ORPHANS_VARIANTS_CLEANUP = Plugin::PREFIX . '/cron/orphans-variant-cleanup';
```
sun, line 48:
```suggestion
  const CRON_EVENT_ORPHAN_VARIANTS_CLEANUP = Plugin::PREFIX . '/cron/orphan-variants-cleanup';
```

src/Plugin.php
```diff
@@ -211,6 +218,13 @@ public static function init() {
       }
     }
 
+    // Schedules delete orphan variants cron.
+    if (!wp_next_scheduled(self::CRON_EVENT_ORPHANS_VARIANTS_CLEANUP)) {
+      wp_schedule_event(time(), 'weekly', self::CRON_EVENT_ORPHANS_VARIANTS_CLEANUP);
+    }
+
```
sun, line 225:
```suggestion
```

src/WooCommerce.php
```diff
@@ -249,6 +249,19 @@ public static function cron_remove_back_in_stock() {
...
+      "DELETE products
+    FROM {$wpdb->posts} products
+    LEFT JOIN {$wpdb->posts} wp ON wp.ID = products.post_parent
+    WHERE wp.ID IS NULL AND products.post_type = 'product_variation';"
+    );
```
sun, line 262: We can't delete the data directly. We actually need to invoke the WordPress API to delete the posts (or WooCommerce product API if there is a wc_delete_product()), so that third-party services like Elasticsearch are updated accordingly.

Realizing this now, this also applies to the core-standards post meta cleanup cron, which should use delete_post_meta(). But maybe it's not that critical in that one, because post meta cannot appear on its own in third-party services.

src/Plugin.php
```diff
@@ -211,6 +218,13 @@ public static function init() {
...
+    add_action(self::CRON_EVENT_ORPHANS_VARIANTS_CLEANUP, __NAMESPACE__  . '\WooCommerce::cron_orphans_variants_cleanup');
```
sun, line 226: Let's move this above the if condition to retain consistency with the other cron setup lines above.

sun (changes requested):

src/Plugin.php
```diff
@@ -40,6 +40,13 @@ class Plugin {
    */
   const CRON_EVENT_REMOVE_BACK_IN_STOCK = Plugin::PREFIX . '/remove-past-back-in-stock';
 
+  /**
+ * Cron event name for delete orphan variants.
```
sun, line 44:
```suggestion
   * Cron event name for delete orphan variants.
```

src/WooCommerce.php
```diff
@@ -249,6 +249,25 @@ public static function cron_remove_back_in_stock() {
...
+  /**
+   * Cron event callback to remove outdated back-in-stock product metadata.
+   */
+  public static function cron_orphan_variants_cleanup() {
+    global $wpdb;
+    $orphan_variants_ids = $wpdb->get_col(
+      "SELECT products.ID products
+      FROM {$wpdb->posts} products
+      LEFT JOIN {$wpdb->posts} wp ON wp.ID = products.post_parent
+      WHERE wp.ID IS NULL AND products.post_type = 'product_variation';"
```
sun, line 261: Did you run this query?  There is a superfluous "products" in the selected fields, which is a syntax/parse error.
```suggestion
      "SELECT p.ID
      FROM {$wpdb->posts} p
      LEFT JOIN {$wpdb->posts} parent ON parent.ID = p.post_parent
      WHERE parent.ID IS NULL AND p.post_type = 'product_variation'"
```

src/WooCommerce.php
```diff
@@ -249,6 +249,25 @@ public static function cron_remove_back_in_stock() {
...
+    );
+
+    if (!empty($orphan_variants_ids)) {
+      foreach ($orphan_variants_ids as $id) {
+        wp_delete_post($id, true);
```
sun, line 266:
```suggestion
        wp_delete_post($id, TRUE);
```

sun: According to https://stackoverflow.com/questions/46874020/delete-a-product-by-id-using-php-in-woocommerce, WooCommerce uses the API of loading the product and invoking the delete method?

```php
$product = wc_get_product($id);
$product->delete(TRUE);
```

src/WooCommerce.php
creative-andrew: @sun , Yes, of course. I never submit coding without running/testing it.  :)

I agree that the products is superfluous, but it doesn't trigger any error: nevertheless 
image

I've been using exactly the same query provided by WooCommerce: https://github.com/woocommerce/woocommerce/blob/3a76aa42555f037a0a565e9f0feaa43f89b197d2/plugins/woocommerce/includes/rest-api/Controllers/Version2/class-wc-rest-system-status-tools-v2-controller.php#L464

src/WooCommerce.php
creative-andrew: Changed to use Woocommerce API

PR #214, opened 2022-08-05
Added customer metafields for mailchimp use.

danielistrate27:
### Ticket
- [Add custom meta fields to orders](https://app.asana.com/0/1201940451841784/1202635304001298)

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -1493,7 +1493,7 @@ class WooCommerce {
   }
 
   /**
-   * Adds custom meta to customer.
+   * Adds customer orders and items count to orders meta.
    *
    * @implements woocommerce_checkout_create_order
    */
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -1517,9 +1517,7 @@ class WooCommerce {
         $items_counter += $customer_order->get_item_count();
       }
 
-      if (!empty($customer_orders)) {
-        update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', count($customer_orders) + 1);
-      }
+      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', count($customer_orders) + 1);
       update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', $items_counter);
     }
   }
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -1503,8 +1503,8 @@ class WooCommerce {
     $customer_orders_count = get_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', TRUE);
     $customer_order_items_count = get_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', TRUE);
     if ($customer_orders_count && $customer_order_items_count) {
-      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', $customer_orders_count + 1);
-      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', $customer_order_items_count + $items_counter);
+      $customer_orders_count++;
+      $customer_order_items_count += $items_counter;
     }
     else {
       $args = [
@@ -1517,9 +1517,11 @@ class WooCommerce {
         $items_counter += $customer_order->get_item_count();
       }
 
-      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', count($customer_orders) + 1);
-      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', $items_counter);
+      $customer_orders_count = count($customer_orders) + 1;
+      $customer_order_items_count = $items_counter;
     }
+    update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', $customer_orders_count);
+    update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', $customer_order_items_count);
   }
 
 }
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -1495,10 +1495,9 @@ class WooCommerce {
   /**
    * Adds custom meta to customer.
    *
-   * @implements woocommerce_checkout_order_processed
+   * @implements woocommerce_checkout_create_order
    */
-  private static function addCustomMetaForUser($order_id) {
-    $order = wc_get_order($order_id);
+  public static function addCustomMetaForUser($order) {
     $user_id = $order->get_user_id();
 
     $args = [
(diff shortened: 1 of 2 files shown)


src/WooCommerce.php
```diff
@@ -1492,4 +1492,32 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+  public static function addCustomMetaForUser($order) {
+    $user_id = $order->get_user_id();
+
+    $args = [
+      'limit' => -1,
+      'customer_id' => $user_id,
+    ];
+    $customer_orders = wc_get_orders($args);
+
+    $items_counter = $order->get_item_count();
```
bogdanarizancu, line 1509: If you initialise `$customer_orders` you don't to check before updating meta:
```suggestion
    $customer_orders = 1;
    $items_counter = $order->get_item_count();
```

src/WooCommerce.php
```diff
@@ -1492,4 +1492,32 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+    foreach ($customer_orders as $o) {
+      $items_counter += $o->get_item_count();
+    }
+
+    if ($customer_orders) {
+      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', count($customer_orders) + 1);
+    }
```
bogdanarizancu, line 1516:
```suggestion
    update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', $customer_orders);
```

bogdanarizancu (changes requested):
@danielistrate27 this is good work, but could we improve this so that it wouldn't always iterate over all past orders if the meta field has already been set once.

src/WooCommerce.php
```diff
@@ -1492,4 +1492,32 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+
+    $args = [
+      'limit' => -1,
+      'customer_id' => $user_id,
+    ];
+    $customer_orders = wc_get_orders($args);
+
+    $items_counter = $order->get_item_count();
+    foreach ($customer_orders as $o) {
+      $items_counter += $o->get_item_count();
```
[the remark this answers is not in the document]
danielistrate27: I changed this in the last commit but I don't think that is ok to use the variable name $order because is already used as the function parameter.

src/WooCommerce.php
danielistrate27: Fixed in the last commits

src/WooCommerce.php
danielistrate27: I can't do this because the woocommerce_checkout_create_order hook is triggered before the save of the current order. So the current order doesn't appear in the query. Thats why we need to get the item count and add a +1 to customer_orders_count field.

src/WooCommerce.php
```diff
@@ -1492,4 +1492,34 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+   * Adds custom meta to customer.
+   *
+   * @implements woocommerce_checkout_create_order
+   */
+  public static function addCustomMetaForUser($order) {
+    $user_id = $order->get_user_id();
+    $items_counter = $order->get_item_count();
+    $customer_orders_count = get_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', TRUE);
+    $customer_order_items_count = get_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', TRUE);
+    if ($customer_orders_count && $customer_order_items_count) {
```
creative-andrew, line 1505: The first time you try to run this     
```
$customer_orders_count = get_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', TRUE);
$customer_order_items_count = get_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', TRUE);
```

it will return an empty string and thus not moving into the conditional.

This will always evaluate false: 

`if ($customer_orders_count && $customer_order_items_count)`

if they user meta is empty for both fields. 

https://developer.wordpress.org/reference/functions/get_user_meta/#return

src/WooCommerce.php
```diff
@@ -1492,4 +1492,34 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
     return $dropdown_values;
   }
 
+  /**
+   * Adds custom meta to customer.
```
creative-andrew, line 1496:
```suggestion
   * Adds customer orders and items count to orders meta.
```

src/WooCommerce.php
```diff
@@ -1492,4 +1492,34 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+   *
+   * @implements woocommerce_checkout_create_order
+   */
+  public static function addCustomMetaForUser($order) {
+    $user_id = $order->get_user_id();
+    $items_counter = $order->get_item_count();
```
creative-andrew, line 1502:
```suggestion
    $items_counter = 0;
```

src/WooCommerce.php
```diff
@@ -1492,4 +1492,34 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+    $customer_orders_count = get_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', TRUE);
+    $customer_order_items_count = get_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', TRUE);
+    if ($customer_orders_count && $customer_order_items_count) {
+      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', $customer_orders_count + 1);
+      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', $customer_order_items_count + $items_counter);
```
creative-andrew, line 1507:
```suggestion
      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_orders_count', $customer_orders_count + 1);
      update_user_meta($user_id, '_' . Plugin::PREFIX . '_customer_order_items_count', $customer_order_items_count + $items_counter);
```

I see this exact code running twice.  Can we DRY it?

creative-andrew (changes requested):

src/WooCommerce.php
danielistrate27: Updated in the last commit. Can you tell me please if this is a good approach?

danielistrate27: I updated this in the last commit. Is this a good approach for DRY?

src/WooCommerce.php
```diff
@@ -1492,4 +1492,36 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
     return $dropdown_values;
   }
 
+  /**
+   * Adds customer orders and items count to orders meta.
+   *
+   * @implements woocommerce_checkout_create_order
+   */
+  public static function addCustomMetaForUser($order) {
```
creative-andrew, line 1500:
```suggestion
  public static function addCustomMetaForUser(WC_Order $order) {
```

src/WooCommerce.php
```diff
@@ -1492,4 +1492,36 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+    $user_id = $order->get_user_id();
```
creative-andrew, line 1501:
```suggestion
    if (!order) {
     return;
    }
```

src/WooCommerce.php
```diff
@@ -1497,7 +1497,10 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
    *
    * @implements woocommerce_checkout_create_order
    */
-  public static function addCustomMetaForUser($order) {
+  public static function addCustomMetaForUser(WC_Order $order):void {
```
creative-andrew, line 1500:
```suggestion
  public static function addCustomMetaForUser(\WC_Order $order):void {
```

creative-andrew: I forgot the slash. Please don't forget to run the code again to see if everything works.

src/WooCommerce.php
creative-andrew: Yes! Thanks.

src/WooCommerce.php
danielistrate27: Updated in the last commit.

creative-andrew (approved):

bogdanarizancu (approved):

PR #215, opened 2022-08-05
Added Diverse option for salutation field.

danielistrate27:
### Ticket
- [Shop Standards: Add third salutation "Divers" ](https://app.asana.com/0/1201940451841784/1201798401981078)


bogdanarizancu (approved):

languages/shop-standards-de_DE.po
```diff
@@ -411,5 +411,9 @@ msgstr "Herr"
 msgid "Company"
 msgstr "Firma"
 
+#: src/WooCommerceSalutation.php:68
+msgid "Divers"
+msgstr "Diverse"
```
sun, line 416: This is reversed - "Diverse" is the English term, "Divers" is the German.  The code should contain the English term.

That said, in English, the term "Other" or "Intersex" may be used more frequently:
https://dict.leo.org/forum/viewUnsolvedquery.php?idThread=1365601&idForum=1&lp=ende&lang=de#followup=4

This is relevant, because the string might have to be translated into other languages in the future.

sun (changes requested):

src/WooCommerceSalutation.php
```diff
@@ -65,6 +65,7 @@ public static function woocommerce_default_address_fields(?array $fields): array
         'Mrs' => __('Mrs', Plugin::L10N),
         'Mr' => __('Mr', Plugin::L10N),
         'Company' => __('Company', Plugin::L10N),
+        'Divers' => __('Divers', Plugin::L10N),
```
sun, line 68: We should probably also move this before Company, otherwise the association with a gender gets lost.

danielistrate27: Updated in the last commit.

languages/shop-standards-de_DE.po
danielistrate27: Updated in the last commit.

PR #217, opened 2022-09-07
Fixed fatal error during checkout if customer does not create account.

sun:
Fixes a fatal error introduced by https://github.com/netzstrategen/wordpress-shop-standards/pull/214 (for [Add custom meta fields to orders](https://app.asana.com/0/1201940451841784/1202635304001298)) ([Slack thread](https://netzstrategen.slack.com/archives/CLGJ41RPW/p1662557855312009?thread_ts=1662553421.058559&cid=CLGJ41RPW))

### Description
- Lots of fatal errors are happening since last deployment on Monday and hundreds of orders were not be able to be placed.
- The new code to track counts of orders per customer assumes that every customer would always create a user account.  But anonymous checkouts are possible.
- This causes the code to load all orders that are associated with user ID 0 (a lot).
- In turn, the server runs out of memory, halting in a fatal error.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -188,8 +188,8 @@ class Plugin {
     // Adds missing postcode validation for some countries.
     add_filter('woocommerce_validate_postcode', __NAMESPACE__ . '\WooCommerce::woocommerce_validate_postcode', 10, 3);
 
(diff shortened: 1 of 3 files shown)


creative-andrew (approved):
Approve with comment.

PR #220, opened 2022-10-24
Added cache invalidation for woocommerce-order-status-manager plugin.

navidabdi:
### Ticket
https://app.asana.com/0/0/1202642566282948/f

### DIFF
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -1514,14 +1514,14 @@ class WooCommerce {
     return $dropdown_values;
   }
 
-   /**
+  /**
    * Invalidates the object cache in Woocommerce order status manager.
-   * 
+   *
    * @implements save_post_wc_order_status
-   * 
    */
   public static function save_post_wc_order_status() {
-    if (wp_cache_get('wc_order_status_manager_order_status_posts')) {
+    wp_cache_get('wc_order_status_manager_order_status_posts', '', FALSE, $found);
+    if ($found) {
       wp_cache_delete('wc_order_status_manager_order_status_posts');
     }
   }
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -1514,14 +1514,16 @@ class WooCommerce {
     return $dropdown_values;
   }
 
-    /**
-   * Invalidate the object cache in Woocommerce order status manager
+   /**
+   * Invalidates the object cache in Woocommerce order status manager.
    * 
    * @implements save_post_wc_order_status
    * 
    */
   public static function save_post_wc_order_status() {
+    if(wp_cache_get( 'wc_order_status_manager_order_status_posts')){
       wp_cache_delete( 'wc_order_status_manager_order_status_posts');
+    }
   }
 
 }
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -27,6 +27,7 @@ class WooCommerce {
    */
   public static function init(): void {
     add_filter(Plugin::PREFIX . '/display_custom_product_fields', __CLASS__ . '::get_product_fields');
+    add_action('save_post', __CLASS__ . '::save_post_custom_status_manager');
   }
 
   /**
@@ -1511,6 +1512,23 @@ class WooCommerce {
     ];
     
     return $dropdown_values;
+  }
+
+    /**
+   * Invalidate the object cache in Woocommerce order status manager
+   * 
+   * @param int $post_id
+   * 
+   * @param \WP_Post $post
+   *   all post types - we're checking for wc_order_status post type
+   * 
+   * @return void
+   * 
+   */
+  private static function save_post_custom_status_manager(int $post_id, \WP_Post $post) {
+    if($post->post_type === 'wc_order_status' && wp_cache_get('wc_order_status_manager_order_status_posts')) {
+      wp_cache_delete( 'wc_order_status_manager_order_status_posts');
+    }
   }
 
 }
--- a/src/WooCommerce.php
+++ b/src/WooCommerce.php
@@ -1517,7 +1517,7 @@ class WooCommerce {
     /**
    * Invalidate the object cache in Woocommerce order status manager
    * 
-   * @return void
+   * @implements save_post_wc_order_status
    * 
    */
   public static function save_post_wc_order_status() {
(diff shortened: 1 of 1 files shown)


src/WooCommerce.php
```diff
@@ -1513,4 +1514,21 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+   * 
+   * @param int $post_id
+   * 
+   * @param \WP_Post $post
+   *   all post types - we're checking for wc_order_status post type
+   * 
+   * @return void
+   * 
+   */
+  private static function save_post_custom_status_manager(int $post_id, \WP_Post $post) {
```
creative-andrew, line 1528: Let's make the function public. 

```suggestion
  public static function save_post_custom_status_manager(int $post_id, \WP_Post $post) {
```

src/WooCommerce.php
```diff
@@ -1513,4 +1514,21 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+    if($post->post_type === 'wc_order_status' && wp_cache_get('wc_order_status_manager_order_status_posts')) {
```
creative-andrew, line 1529: I think we can use 
https://developer.wordpress.org/reference/hooks/save_post_post-post_type/
to target specifically that post type and that will make the if statement obsolete.

src/WooCommerce.php
```diff
@@ -1513,4 +1514,14 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
     return $dropdown_values;
   }
 
+    /**
+   * Invalidate the object cache in Woocommerce order status manager
+   * 
+   * @return void
```
creative-andrew, line 1520:
```suggestion
   * @implements save_post_wc_order_status
```

src/WooCommerce.php
```diff
@@ -1513,4 +1514,14 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
     return $dropdown_values;
   }
 
+    /**
+   * Invalidate the object cache in Woocommerce order status manager
+   * 
+   * @implements save_post_wc_order_status
+   * 
+   */
```
creative-andrew, line 1522:
```suggestion
  /**
   * Invalidates the object cache in Woocommerce order status manager.
   * 
   * @implements save_post_wc_order_status
   * 
   */
```

src/WooCommerce.php
```diff
@@ -1513,4 +1514,14 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
+  public static function save_post_wc_order_status() {
+      wp_cache_delete( 'wc_order_status_manager_order_status_posts');
```
creative-andrew, line 1524: I think we should check like you did before if we have someone in the cache first, and if we do we can delete it.

src/WooCommerce.php
```diff
@@ -1514,14 +1514,16 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
-    /**
-   * Invalidate the object cache in Woocommerce order status manager
+   /**
+   * Invalidates the object cache in Woocommerce order status manager.
    * 
    * @implements save_post_wc_order_status
    * 
    */
   public static function save_post_wc_order_status() {
+    if(wp_cache_get( 'wc_order_status_manager_order_status_posts')){
```
creative-andrew, line 1524:
```suggestion
    if (wp_cache_get('wc_order_status_manager_order_status_posts')) {
```

src/WooCommerce.php
```diff
@@ -1514,14 +1514,16 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
       wp_cache_delete( 'wc_order_status_manager_order_status_posts');
```
creative-andrew, line 1525:
```suggestion
      wp_cache_delete('wc_order_status_manager_order_status_posts');
```

src/WooCommerce.php
```diff
@@ -1513,4 +1514,16 @@ private static function getRepricingOptionsDropdownValues(\WC_Product $product)
...
   }
 
+   /**
+   * Invalidates the object cache in Woocommerce order status manager.
+   * 
+   * @implements save_post_wc_order_status
+   * 
+   */
+  public static function save_post_wc_order_status() {
+    if (wp_cache_get('wc_order_status_manager_order_status_posts')) {
```
sun, line 1524: 1. Is there a wp_cache_has() that only looks up but does not fetch the cache item? If not, please use the `$found` argument of wp_cache_get() to determine whether the cache item was found.

2. This fix is so trivial, why aren't we contributing it as a patch to the plugin vendor (upstream PR if possible?) and temporarily apply it via shared-patches to the actual plugin instead?  Otherwise we'll have to remove this code fairly soon again once upstream is fixed?

sun (commented):
Hey, sorry for the delay. Here's my review.

src/WooCommerce.php
navidabdi: Hi, @sun, Thank you for your great review, 
I searched and the method wp_cache_has() does not exist, and also can you explain what you mean:
**use the $found argument of wp_cache_get() to determine whether the cache item was found.**

creative-andrew: `if (false === wp_cache_get('wc_order_status_manager_order_status_posts'))`

sun: Just recognized that this is indeed not noted in the manual yet, so I explained it here: https://developer.wordpress.org/reference/functions/wp_cache_get/#comment-6157

The post is pending moderation approval, so here's a copy:

---
Here is how you use the `$found` parameter in your code – it is passed by reference, so you do not need to define the variable upfront, its value will be set by the called function instead:

```php
$value = wp_cache_get('mykey', 'mygroup', false, $found);
if (!$found) {
  $value = my_expensive_function();
  wp_cache_set('mykey', $value, 'mygroup');
}
return $value;
```
You need to use the `$found` parameter if there is a slightest chance that the value you are caching can be zero, false, null, or otherwise equivalent to false like an empty array.

If you are not sure, or if you are looking for **a general rule of thumb – always use the `$found` parameter.**

If some object caches are not implementing the `$found` parameter correctly, then that is a good reason to fix those implementations (or use better maintained ones) and not a reason to not use the parameter.

sun: Did you work with https://github.com/makers99/wp-cli-shared-patches already?

navidabdi: > 

Thank you so much for your awesome clarification 👍
Unfortunately, I didn't work with  https://github.com/makers99/wp-cli-shared-patches yet.

sun:
Replaced by https://github.com/makers99/wp-cli-shared-patches/pull/32 and https://github.com/netzstrategen/wopa/pull/1132

PR #231, opened 2024-11-18
feat: exclude product from comparison platforms checkbox

creative-andrew:
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208758137825995


navidabdi (approved):
LGTM :-)

PR #232, opened 2024-11-20
fix: warning undefined notice

creative-andrew:
image

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208770722582157

### DIFF
--- a/src/Admin.php
+++ b/src/Admin.php
@@ -114,7 +114,8 @@ class Admin {
 
     foreach ($product->get_children() as $variation) {
       $variation_term_id = get_post_meta($variation, '_lieferzeit', TRUE);
-      $variation_term_slug = get_term($variation_term_id)?->slug;
+      $variation_term = get_term($variation_term_id);
+      $variation_term_slug = $variation_term->slug;
       if (!$variation_term_slug) {
         continue;
       }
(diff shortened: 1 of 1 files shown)


src/Admin.php
```diff
@@ -114,7 +114,10 @@ public static function getProductShortestDeliveryTime(\WC_Product $product): str
 
     foreach ($product->get_children() as $variation) {
       $variation_term_id = get_post_meta($variation, '_lieferzeit', TRUE);
-      $variation_term_slug = get_term($variation_term_id)->slug;
+      $variation_term_slug = get_term($variation_term_id)?->slug;
+      if (!$variation_term_slug) {
+        continue;
+      }
```
sun, line 120: Cleaner:
```suggestion
      if (!$variation_term = get_term($variation_term_id)) {
        continue;
      }
      $variation_term_slug = $variation_term->slug;
```

sun (approved):
Thanks!  Please fix the PR title to properly state what problem is fixed before merging.

src/Admin.php
creative-andrew: I don't like assigning stuff inside if statements at all. But thanks for the suggestion

PR #246, opened 2026-07-08
Add collapsible product attributes, price labels and checkout fields for the Cart/Checkout blocks

creative-andrew:
### Ticket
- N/A — multi-brand consolidation of block cart/checkout work (WOPA/LAMP/GACO all run this stack)

### Description

The Cart and Checkout blocks render line-item metadata and checkout steps client-side from the Store API, so the classic cart-item-data.php template override and classic checkout-field hooks don't apply there. This PR adds everything needed to support that in the blocks:

- **Server-side tagging**: hides the synthetic `separator` item-data row in the blocks (it's only meaningful to the classic template) and tags each product-attribute row with a `className` the block copies onto its `<li>`, giving the client script a stable hook.
- **Collapsible product attributes** in the Cart block and Checkout order summary, matching the classic cart's toggle, via a new `WooCommerceBlocks` integration.
- **Cart line-item price labels** (`Einzelpreis (UVP)`/`(Sale)`, `Gesamt:`, per-unit/cumulated savings qualifiers, one-line MwSt. note) and removal of German Market's "Veranschlagte" prefix from the block order total.
- **Block checkout fields**: salutation (gated on the existing `shop-standards_add_salutation_field` option, shared options with the classic field via `WooCommerceSalutation::getSalutationOptions()`), academic title, and — for MOEVE stores — a floor-number field, all mirrored to the classic `_billing_*`/`_shipping_*` meta so existing consumers (MOEVE's order export, etc.) need no changes.
- **MOEVE's "company name required" validation** re-implemented for the block checkout (a "Company" salutation requires a company name).
- **Conditional EU VAT field**: the EU VAT Number plugin's VAT step is shown only for EU companies outside Germany; the plugin itself is untouched, so VIES validation and reverse-charge stay intact.

Built and verified live against WooCommerce 10.5.1. Dist assets were built with `npx sass`/`npx terser` since the project's gulp/node-sass toolchain doesn't build on current Node versions — re-buildable with `npm run build` once that's resolved; functionally identical output.


assets/styles/blocks.scss
```diff
@@ -0,0 +1,368 @@
...
+  box-sizing: border-box !important;
+  width: auto !important;
+  height: auto !important;
+  min-width: auto !important;
+  min-height: auto !important;
+  margin: 0 !important;
+  padding: 0.5em 1em !important;
+  border: 2px solid var(--c-brand, #143264) !important;
+  border-radius: 5px !important;
+  background: var(--c-brand, #143264) none !important;
```
sun: It seems these CSS variables are custom to WOPA/GACO, so they should not be placed in the plugin.

assets/styles/blocks.scss
```diff
@@ -0,0 +1,368 @@
...
+  .wc-block-components-address-form__last_name { order: 5; }
+  .wc-block-components-address-form__address_1 { order: 6; }
+  .wc-block-components-address-form__address_2-toggle { order: 7; }
+  .wc-block-components-address-form__address_2-hidden-input { order: 8; }
+  .wc-block-components-address-form__postcode { order: 9; }
+  .wc-block-components-address-form__city { order: 10; }
+  .wc-block-components-address-form__state { order: 11; }
+  .wc-block-components-address-form__phone { order: 12; }
+  .wc-block-components-select-input-shop-standards-floor-number { order: 13; }
+}
```
sun: The address form fields and their order are different per country, because every country has its own habit.  The placement of additional fields should be controlled via the PHP hooks.

assets/styles/blocks.scss
```diff
@@ -0,0 +1,368 @@
...
+.wc-block-cart-item__wrap > .wc-block-components-sale-badge::after {
+  content: ' pro Stück';
+}
+
+// Right column (cumulated): label the line total, keep the tax note on one
+// line, and mark the cumulated savings badge.
+.wc-block-cart-item__total {
+  .wc-block-components-product-price__value::before {
+    content: 'Gesamt: ';
+  }
```
sun: These do not work for sites that are not in German language.

dist/scripts/block-labels.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Adjusts Cart/Checkout block labels: via the checkout filter registry where
+ * a filter exists, otherwise via direct DOM text updates.
+ */
+(function cartBlockLabels(wp, wc) {
+  if (wc && wc.blocksCheckout && typeof wc.blocksCheckout.registerCheckoutFilters === 'function') {
+    wc.blocksCheckout.registerCheckoutFilters('shop-standards-block-labels', {
+      // Removes German Market's "Veranschlagte" prefix from the block total.
+      totalLabel: (value) => (typeof value === 'string' ? value.replace(/^Veranschlagte\s+/i, '') : value),
```
sun: This is actually coming from WooCommerce Core.

dist/scripts/block-vat.js
```diff
@@ -0,0 +1,111 @@
...
+  const SALUTATION_FIELD = 'shop-standards/salutation';
+
+  // Rewords the step heading from "VAT Details" to "Umsatzsteuer-ID".
+  function fixHeadingText() {
+    document.querySelectorAll(VAT_HEADING).forEach((el) => {
+      if (el.textContent.trim() === 'VAT Details') {
+        el.textContent = 'Umsatzsteuer-ID';
+      }
+    });
+  }
```
sun: Step headings can be configured with the FSE for the checkout page

dist/scripts/block-vat.js
```diff
@@ -0,0 +1,111 @@
...
+   */
+  function shouldShow() {
+    const store = wp.data.select('wc/store/cart');
+    if (!store) {
+      return false;
+    }
+    const billing = store.getCartData().billingAddress || {};
+    const country = billing.country || '';
+    const salutation = billing[SALUTATION_FIELD] || '';
+    return country !== '' && country !== 'DE' && salutation === 'Company';
```
sun: Embedding the value "company" into a salutation field is not something I would recommend to do and therefore would not add to the shared shop-standards plugin.

If you need it for WOPA/GACO specifics, then I would rather suggest to create a dedicated plugin that can be shared between them.

creative-andrew: I will create a new plugin as these are really specific business rules for GACO/WOPA/LAMP.

assets/styles/blocks.scss
creative-andrew: Unfortunately, I believe that is no longer the case with the new block checkout. 
The new cart and checkout blocks offer a lot less flexibility for customization and most of it cannot be controlled via PHP.

creative-andrew: Thanks for flagging this — wanted to verify before replying, so I traced it through both code and a live test rather than just reasoning about it.

**You're right that WooCommerce has per-country address formats.** `WC_Countries::get_address_formats()` in core defines genuinely different layouts per country — e.g. default is `{city}\n{state}\n{postcode}`, while Japan is `{postcode}\n{state} {city} {address_1}` and even swaps first/last name order. That's real.

**But that format data doesn't drive the editable form fields — it drives the collapsed step summary.** I traced `defaultAddressFormat` / `countryData[country].format` through the built JS and it's consumed by the read-only "✓ step completed, click to edit" recap component (`onEdit`/`isExpanded` props), not `.wc-block-components-address-form`. The actual editable form appears to use a fixed field-key list (`first_name, last_name, company, address_1, address_2, city, postcode, country, state, phone`) regardless of country.

I confirmed this empirically too: with this CSS temporarily disabled, I switched the live checkout's country between DE/FR/AT (everything this store ships to) and diffed the native DOM order each time. It was identical in all three — only whether `state` is present changes (hidden for FR/AT), never the relative order of the other fields.

I also checked whether "controlled via PHP hooks" is actually possible right now: `CheckoutFields::register_checkout_field()`'s accepted options are `id, label, location, type, hidden, required, attributes, ...` — there's no position/order/priority param, and new fields are simply appended to their location group in registration order (`$this->fields_locations[$location][] = $field_id`). So there's currently no WooCommerce Blocks hook that can control field position — CSS `order` is the only lever the API exposes today.

Net: for this store's actual shipping zone (DE/FR/LI/AT/CH — all the same postal convention), I don't have evidence the field order varies, and there's no PHP-level alternative available. The valid long-term concern is if this store ever ships to a country with a genuinely different convention (US/UK/JP) — worth revisiting then, possibly by mirroring how the step-summary component reads `country.format` dynamically. Happy to add a code comment noting this limitation if that's useful, but I don't think there's a more correct implementation available right now given what WooCommerce Blocks exposes.

creative-andrew:
Following up on the review discussion about the checkout-field business rules (salutation/title/floor, conditional VAT, company-name validation) not belonging in a generic, shared plugin — moved them out into a new plugin, [netzstrategen/shop-commons](https://github.com/netzstrategen/shop-commons), business-specific to WOPA/GACO/LAMP.

Removed from this PR in cbae6d5: `WooCommerceCheckoutBlockFields.php`, `block-vat.js`, and the checkout-field/VAT-step CSS. `shop-standards` now only keeps the generic Cart/Checkout block work (collapsible attributes, price labels, buttons, advantages, Trusted Shops banner).

Verified live on wopa.test with both plugins active: fields render in the same position with no duplication, VAT step shows/hides correctly, no fatal errors.

creative-andrew:
Following up: all of the Cart/Checkout block work on this branch — the checkout fields, and also the collapsible attributes, price/sale-badge labels, brand buttons, shop advantages, and Trusted Shops banner — has moved to a new business-specific plugin, [netzstrategen/shop-commons](https://github.com/netzstrategen/shop-commons), shared by WOPA/GACO/LAMP.

`shop-standards`' branch has been reverted to match `master` exactly, since nothing from it remains here. Closing this PR without merging — there's nothing left in `shop-standards` to land.

